PR #2, opened 2024-05-15
Changes to get terraform project deployed through spacelift

BryanFauble:
**Problem:**

1. We needed to be able to deploy cloud infra outside of a k8s cluster and within a k8s cluster

**Solution:**

1. Using some pre-built terraform modules to accomplish this goal
2. Moving to a helm terraform provider to deploy airflow instance

**Testing:**

1. Tested that the k8s cluster has the helm chart installed, spotio is managing scaling, airflow can run and may access the appropriate secrets

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -126,31 +126,31 @@ module "eks" {
   #   instance_types = ["m6i.large", "m5.large", "m5n.large", "m5zn.large"]
   # }
 
-  # eks_managed_node_groups = {
-  #   one = {
-  #     name         = var.eks_nodeGroup
-  #     desired_size = 1
-  #     min_size     = 0
-  #     max_size     = 10
+  eks_managed_node_groups = {
+    one = {
+      name         = var.eks_nodeGroup
+      desired_size = 1
+      min_size     = 0
+      max_size     = 10
 
-  #     instance_types = ["t3.large"]
-  #     capacity_type  = "SPOT"
-  #     iam_role_additional_policies = {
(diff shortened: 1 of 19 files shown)


main.tf
```diff
@@ -126,31 +126,31 @@ module "eks" {
...
-  #   #   instance_types = ["t3.large"]
-  #   #   capacity_type  = "SPOT"
-  #   # }
-  # }
+  eks_managed_node_groups = {
+    one = {
+      name         = var.eks_nodeGroup
+      desired_size = 1
+      min_size     = 0
+      max_size     = 10
```
thomasyu888, line 134: Spot recommended us to have these
```suggestion
      desired_size = 0
      min_size     = 0
      max_size     = 0
```

BryanFauble: I swapped this to a max_size of 1 (0) is not valid.

Where did they recommend this? Their docs are all over the place and I couldn't find this piece

main.tf
```diff
@@ -122,80 +114,61 @@ module "eks" {
...
+            type = "cluster"
+          }
+        }
+      }
+    }
+  }
   tags = var.tags
 }
 
+module "ocean-controller" {
```
BryanFauble, line 145: @thomasyu888 Something I could use some help on. I wasn't clear based on their published docs which modules I was really supposed to use here. Specifically the issue I am seeing in the Spot console is:


```
Different controller versions detected. Please remove the old version controller. Old version controller pod(s): [spotinst-kubernetes-cluster-controller-685b9885bb-7nll9], New version controller pod(s): [ocean-controller-ocean-kubernetes-controller-77b5cb4c7-6dt4g, ocean-controller-ocean-kubernetes-controller-77b5cb4c7-wbslc] 
```

image

PR #4, opened 2024-05-20
[IBCDPE-939] Correcting some issues

BryanFauble:
**Problem:**

1. No instructions were provided on authenticating to the cluster
2. We were not setting a version on the terraform provider. I set this to the max provided by spacelift. To go beyond this version we'll need to consider using the OpenTofu backend/provider.
3. We didn't have a CODEOWNERS
4. Permissions should not have been given as administrator
5. spotinst has 3 item currently being deployed. From the documentation I can find `spotinst/kubernetes-controller/ocean` seems like it was a mistake to add this
6. We had no automatic checking of terraform in the CI pipeline

**Solution:**

1. Adding all of the above items

**Testing:**

1. Will be checking the terraform plan

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+* @Sage-Bionetworks-Workflows/dpe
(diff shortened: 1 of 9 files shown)


BryanFauble:
Creating another PR so the changes show in spacelift

PR #8, opened 2024-05-28
[IBCDPE] Split out and start creating items as individual modules

BryanFauble:
Problem:

1. Spotocean controller V2 is not ready, use are using it wrong, or the documentation is incorrect.
2. We need to start thinking about how we'll organize both modules going into stacks, and the spacelift stacks themselves.

Solution:

1. Revert to Spotocean controller v1
2. Start splitting out the spotocean autoscaler to it's own module
3. Split out airflow into it's own module
4. Start to create some of the scaffolding for programmatically creating spacelift resources like policy, stack, and policy attachment

Testing:

1. Will be verifying that I am able to register these modules in the spacelift terraform registry
2. Will verify that the spot ocean autoscaler correct deploys


modules/internal-k8-infra/main.tf
thomasyu888, line 1: Do you have a list of questions you'd like spot to answer?

stacks/dpe-prod/main.tf
thomasyu888, line 1: Are we currently using these modules?

main.tf
```diff
@@ -1,174 +1,159 @@
...
+  # To add the current caller identity as an administrator
+  enable_cluster_creator_admin_permissions = true
+  authentication_mode                      = "API"
+
+
+  access_entries = {
+    # One access entry with a policy associated
+    eks_admin_role = {
+      kubernetes_groups = []
+      principal_arn     = "arn:aws:iam::766808016710:role/eks_admin_role"
```
thomasyu888, line 158: Could we reference this role?
```
# Not sure if this works
aws_iam_role.admin_role.name
```

main.tf
```diff
@@ -1,174 +1,159 @@
...
+  eks_managed_node_group_defaults = {
+    instance_types = ["m6i.large", "m5.large", "m5n.large", "m5zn.large"]
+  }
+
+  eks_managed_node_groups = {
+    one = {
+      name         = var.eks_nodeGroup
+      desired_size = 1
+      min_size     = 0
+      max_size     = 2
```
thomasyu888, line 123: This is still something I'd like to figure out, spot claimed on the call that we didn't need a managed node group.

thomasyu888 (approved):
🔥 Thanks for working on this - just going to pre-approve, I just have minor comments.

main.tf
BryanFauble: So the problem is that in order for the add ons to be installed in the first place EKS has to have at least 1 node running as that is where the tools are installed to. This means that in order for the EKS cluster to be "green" it has to have at least a single node.

After this point we can install spot ocean.

Even their docs here: https://github.com/spotinst/terraform-spotinst-ocean-eks?tab=readme-ov-file#following-modules-should-be-used-instead--

Specify that we should:

1. Use the AWS EKS provider to create an EKS cluster and it's resources
2. Then install the ocean controllers/k8s

stacks/dpe-prod/main.tf
BryanFauble: No. This is planned future work to start programmatically creating the entire stack within spacelift

spacelift/modules/policy-attachments/main.tf
```diff
@@ -0,0 +1,21 @@
+resource "spacelift_policy_attachment" "bfauble-enforce-tags-on-resources" {
```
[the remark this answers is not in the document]
BryanFauble: When we tear down the existing spacelift stacks and move to a totally programmatic approach yes. I named the first stack 'bfauble', and even if you update the display name the ID of the stack is sticky and does not change.

modules/internal-k8-infra/main.tf
BryanFauble: 1. What terraform modules are required to install the V2 spot ocean components into our EKS cluster?
2. What dependencies need to be completed, and operational before attempting to install any spot ocean terraform resources?
3. What does the removal of ocean's terraform resources entail, is everything being properly reset back to the "before" state if removed?

stacks/dpe-prod/main.tf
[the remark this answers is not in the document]
BryanFauble: I left this in as scaffolding for how I am thinking we'll organize our spacelift resources. Basically we need:

1. Modules that define what we can create
2. Another set of resources that combine those modules in a meaningful way (ie: This is what I am intending this `dpe-prod/main.tf` to do. This is where we create the individual "stacks" we'll be deploying out.
3. Another set of spacelift specific resources (Like policies, stacks, contexts) and their relationships. When a spacelift stack is created it requires that you are pointing at a specific VCS location where it's pulling terraform resources from. That VCS location is defining the resources that "this" spacelift stack i going to deploy.

PR #10, opened 2024-06-11
Update default storage to be gp3

BryanFauble:
https://sagebionetworks.jira.com/browse/IT-2871

Update default storage to be gp3


thomasyu888 (approved):
💸 LGTM!

PR #11, opened 2024-07-11
[IBCDPE-935] Setting up declerative defintion of TF resources

BryanFauble:
**Problem:**

1. Creation and management of resources within spacelift was largely manual setting values in the UI and a "click-ops" procedure.
2. Organization of the TF resources was needed.

**Solution:**

1. Managing the deployment of all resources both spacelift and AWS declarative and automatically with auto-deploy.
2. Using https://github.com/antonbabenko/terraform-best-practices/tree/master/examples as inspiration for directory structure, as well as the great work that @philerooski did for the snowflake tables in RECOVER.


dev/stacks/dpe-sandbox-k8s-deployments/main.tf
```diff
@@ -0,0 +1,327 @@
...
+        }
+      }
+
+      spec {
+        container {
+          name              = "backend"
+          image             = "calico/star-probe:v0.1.0"
+          image_pull_policy = "Always"
+
+          command = ["probe", "--http-port=6379", "--urls=http://frontend.stars:80/status,http://backend.stars:6379/status,http://client.client:9000/status"]
```
thomasyu888, line 243: Nice, I'll have to read into calico more but looks like a good open source tool for networking

main.tf
```diff
@@ -1,181 +1,53 @@
...
-    # AmazonEKSAdminViewPolicy?
-    # AmazonEKSEditPolicy
-    # AmazonEKSViewPolicy
-
-  }
-  tags = var.tags
-}
-
+import {
+  # The initial administrative stack is created manually in the Spacelift UI, and imported
```
thomasyu888, line 2: Is that because it's just easier to create manually in spacelift UI?

modules/main.tf
```diff
@@ -0,0 +1,62 @@
+resource "spacelift_module" "sage-aws-vpc" {
+  github_enterprise {
+    namespace = "Sage-Bionetworks-Workflows"
+    id        = "sage-bionetworks-workflows-gh"
+  }
+
+  name               = "sage-aws-vpc"
+  terraform_provider = "aws"
+  administrative     = false
+  branch             = "ibcdpe-935-vpc-updates"
```
thomasyu888, line 10: Are you thinking we would copy `dev` and have a `prod` when we get there?  Or would we parametrize {branch} based off of the branch we are working on and have `dev` and `prod` set up as blue green deployments

dev/spacelift/dpe-sandbox/main.tf
```diff
@@ -0,0 +1,125 @@
...
+
+resource "spacelift_aws_integration_attachment" "k8s-aws-integration-attachment" {
+  integration_id = "01HXW154N60KJ8NCC93H1VYPNM"
+  stack_id       = spacelift_stack.k8s-stack.id
+  read           = true
+  write          = true
+}
+
+resource "spacelift_aws_integration_attachment" "k8s-deployments-aws-integration-attachment" {
+  integration_id = "01HXW154N60KJ8NCC93H1VYPNM"
```
thomasyu888, line 121: should this integration_id be a secret?

thomasyu888 (approved):
🔥 LGTM!  Going to pre-approve.  I really like the structure, just have some questions about future state in terms of CI/CD of dev/prod and where you envision this going.

dev/spacelift/dpe-sandbox/main.tf
```diff
@@ -0,0 +1,125 @@
...
+
+resource "spacelift_stack" "k8s-stack" {
+  github_enterprise {
+    namespace = "Sage-Bionetworks-Workflows"
+    id        = "sage-bionetworks-workflows-gh"
+  }
+
+  administrative          = false
+  autodeploy              = true
+  branch                  = "ibcdpe-935-vpc-updates"
```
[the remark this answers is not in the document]
BryanFauble: Yes - I'll be swapping these over to `main`

dev/spacelift/dpe-sandbox/main.tf
```diff
@@ -0,0 +1,125 @@
...
+  description             = "Infrastructure to support deploying to an EKS cluster"
+  name                    = "DPE DEV Kubernetes Infrastructure"
+  project_root            = "dev/stacks/dpe-sandbox-k8s"
+  repository              = "eks-stack"
+  terraform_version       = "1.7.2"
+  terraform_workflow_tool = "OPEN_TOFU"
```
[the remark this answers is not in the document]
BryanFauble: That's correct. Spacelift has not updated to newer version of hashicorp terraform at this point in time. So if we want to use some newer modules folks have put out (Like some of the new spot.io modules) we need to be on a newer version.

1.8 of opentofu is around the corner and they're moving steadily in the right direction to make working with TF resources even easier. I am really hopeful for the future of the project.

dev/stacks/dpe-sandbox-k8s-deployments/main.tf
BryanFauble: In this case I am only using these k8 resources to easily prove out some networking within the cluster. At this point in time I am not planning to implement custom networking; instead am going to continue using the AWS VPC-CNI plugin for EKS. There is more tweaking I'll need to make to get all of it to work like we want.

There are 2 ways AWS suggests to handle networking:

1. There is pod level security groups - Meaning each pod can be assigned an AWS security group, and we get very fine grained control over it. So we can easily limit pod -> pod connections, pod -> AWS resources, pod -> Internet, Internet -> pod connections
2. They also allow us to use a k8 resource `NetworkPolicy`, but I need to spend more time to understand how it works with security groups & if both as needed; What the best practices are.

main.tf
BryanFauble: The recommendation from spacelift is to make the first admin stack in the UI and then use that to create the rest of the spacelift resources. It's possible to create this on my local machine, but it was a lot easier just to make the stack, point it to this repo, branch, and directory - then let it work.

All I needed to do at that point was to import it into the terraform state so the admin stack itself is managed by itself.

modules/main.tf
BryanFauble: In the docs they talk a little bit about it here:
https://docs.spacelift.io/integrations/source-control/github#multi-stack-version

> One frequent type of setup involves two similar or even identical environments - for example, staging and production. One approach would be to have them in a single repository but in different directories, setting [project_root](https://docs.spacelift.io/concepts/configuration/runtime-configuration#project_root-setting) runtime configuration accordingly. This approach means changing the staging directory a lot and using as much or as little duplication as necessary to keep things moving, and a lot of commits will necessarily be no-ops for the production stack. This is a very flexible approach, and we generally like it, but it leaves Git history pretty messy and some people really don't like that.

I was thinking of not using long-lived branches for this project (No dev, staging, branch ect...) and instead handle it by moving resources between the appropriate folders are we moved items up. I do recognize that also brings some complexity and possibility to leave behind changes if they're not moved/copied to the appropriate locations.

Conversely long-lived branches has similar issues too where 1 branch can be wildly out-of-date; You have conflicts between branches where you cannot merge dev into staging (for example), you have to merge your feature branch into dev and staging. Which that could also bring problems where a feature branch doesn't get moved to production for a long time, and dev/staging are in an awkward state.


This is all to say - I haven't fully settled on one or the other.

What does @Sage-Bionetworks-Workflows/dpe think on this matter?

dev/spacelift/dpe-sandbox/main.tf
BryanFauble: No this is not a secret. This is the ID in spacelift of the AWS integration for our account - It's an area I haven't nailed down how we'll handle programmatically, so for now I was just going to create the AWS integrations in the UI, and reference the IDs in the TF files.

modules/main.tf
BryanFauble: This talks a bit about trunk based development:
https://trunkbaseddevelopment.com/

BryanFauble: @thomasyu888 Here is that blog post that I finally found:

https://medium.com/env0/terraform-files-and-folder-structure-organizing-infrastructure-as-code-749572469d04

They show using some structures like:

image


Which is close to what we are doing here, instead, we are just putting the dev/prod/staging into the root directory.

The thing though is that it does not cover what moving from dev -> staging looks like :(

BryanFauble: Something else I was thinking about this as well. One way we could go about this is that we do not put any environment specific logic within the stacks we are building. So rather than having an environment specific folder, and then a replication of all the TF resource files we:

1. Create a spacelift `context` for each of the environments that we are going to be running the stacks in
2. Create spacelift `stacks` for each of those environments so they're all handled with different CICD processes
3. Have 1 definition of the cloud resources that we apply to each environment with variables coming
4. If we need to have any environment specific logic we can conditionally create it based on the variables defined in the context

main.tf
thomasyu888: Thanks, could that be added as a comment?

modules/main.tf
thomasyu888: I like that idea.  That is sort of the way sceptre has it to have configurations per environment and then having central infrastructure stacks that aren't duplicated.

PR #12, opened 2024-07-18
[IBCDPE-935] private worker pool module

BryanFauble:
**Problem:**

1. I wanted to take a little time and explore this enterprise feature while we had trial access. Private workers allow us to use drift detection for spacelift stacks.

**Solution:**

1. I created a terraform module for us to use that deploys private workers to a k8s cluster. I added notes in the readme about it as well so we can pick it up again in the future if we want to use this feature.

**Testing:**

1. I was able to deploy the private workers and turn on drift detection. See #dpe-topics slack channel.

### DIFF
--- a/common-resources/policies/main.tf
+++ b/common-resources/policies/main.tf
@@ -1,18 +1,24 @@
 resource "spacelift_policy" "enforce-tags-on-resources" {
-  name = "Enforce Tags On Resources - cli"
-  body = file("${path.module}/enforce-tags-on-resources.rego")
-  type = "PLAN"
-  labels = ["compliance", "plan", "tagging", "terraform"]
+  name        = "Enforce Tags On Resources - cli"
+  body        = file("${path.module}/enforce-tags-on-resources.rego")
+  type        = "PLAN"
+  labels      = ["compliance", "plan", "tagging", "terraform"]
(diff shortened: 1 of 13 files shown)


common-resources/policies/main.tf
```diff
@@ -1,18 +1,24 @@
...
-  body = file("${path.module}/check-estimated-cloud-spend.rego")
-  type = "PLAN"
+  name     = "Cloud Spend Estimation - cli"
+  body     = file("${path.module}/check-estimated-cloud-spend.rego")
+  type     = "PLAN"
   space_id = "root"
 }
 
 
+resource "spacelift_policy" "drift-detection-warning" {
```
thomasyu888, line 19: If we don't have the premium policy , does this work?

thomasyu888 (approved):
🔥 LGTM!  Just a small comment in the code

common-resources/policies/main.tf
BryanFauble: Defining policies in this file just makes them available to be used. It does not add it to any stacks.

If this policy were added to a stack it would always pass because of the condition in `.rego` file:
`input.spacelift.run.drift_detection`

`drift_detection` would always not exist or be false in all cases.

Since I had already written this I was just going to leave it in.

PR #14, opened 2024-07-29
[IBCDPE-1007] Monitoring and security scanning

BryanFauble:
**Problem:**

1. When running workloads on the K8s cluster we only had a small windows to view what was going on inside of the cluster.

**Solution:**

1. Implementing Victoria Metrics (Similar to Prometheus) that allows us to scrape metrics on the K8s cluster.
2. Implementing trivy to handle security and vulnerability scanning/best practices. This operator also creates new scans as resources are added to the cluster giving updates to the results. It also produces new scans every 6 hours.
3. Adding policy-reporter to give a nice UI on top of the scan reports

**Testing:**

1. I was able to deploy all of this out into the dev cluster and verify that as changes are made to the cluster new scans are produced.
2. I also verified that I could add new scrape configs and view the data in grafana dashboards.


modules/sage-aws-k8s-node-autoscaler/main.tf
```diff
@@ -102,6 +102,8 @@ module "ocean-aws-k8s" {
   is_aggressive_scale_down_enabled = true
   max_scale_down_percentage        = 33
   tags                             = var.tags
+  # TODO: Fix this it does not seem to work
```
thomasyu888, line 105: Weird - what doesnt seem to work about this?

modules/trivy-operator/README.md
```diff
@@ -0,0 +1,39 @@
+# Purpose
+
```
thomasyu888, line 2: Just a note: I really like trivy and that would be a nice addition to our GitHub Ci/CD around scanning of docker repos and we build.

Great addition!

dev/spacelift/dpe-sandbox/main.tf
```diff
@@ -13,7 +13,7 @@ resource "spacelift_stack" "k8s-stack" {
 
   administrative          = false
   autodeploy              = true
-  branch                  = "main"
+  branch                  = "ibcdpe-1007-monitoring"
```
thomasyu888, line 16: I might have said this before but it would be nice if there was a parameter to get the branch

modules/main.tf
```diff
@@ -70,35 +117,45 @@ resource "spacelift_module" "spacelift-private-workerpool" {
...
-  administrative     = false
-  branch             = "ibcdpe-935-vpc-updates"
-  description        = "Module for the spacelift private workerpool helm chart which deploys the K8s operator"
-  repository         = "eks-stack"
-  project_root       = "modules/spacelift-private-worker"
-  space_id           = "root"
+  name               = each.value.name
+  terraform_provider = each.value.terraform_provider
+  administrative     = each.value.administrative
+  branch             = each.value.branch
```
thomasyu888, line 147: Can this be used instead of hard coding the branch name (see comment above)

thomasyu888 (commented):
Adding initial thoughts - but really awesome work here!

modules/sage-aws-k8s-node-autoscaler/main.tf
BryanFauble: I had assumed this would force the cluster to scale up the number of nodes (I wanted to check what the behavior was). But it seemed to stay at a single node. I'll update the comment.

dev/spacelift/dpe-sandbox/main.tf
BryanFauble: Yeah I need to figure out a more sustainable way. If we have multiple people working on the repo it's going to be messy.

I think once I get this PR and the VPC PR merged into main, that using trunk based development is going to be a bit nicer. Unfortunately since I am moving faster than code can be reviewed, this was the most simple thing for the mass amount of changes being implemented.

modules/main.tf
BryanFauble: Yes - This is the approach I am planning when I clean up the definition of tf resources

dev/stacks/dpe-sandbox-k8s-deployments/main.tf
```diff
@@ -1,10 +1,27 @@
...
+  source  = "spacelift.io/sagebionetworks/victoria-metrics/aws"
+  version = "0.0.7"
+}
+
+module "trivy-operator" {
+  source  = "spacelift.io/sagebionetworks/trivy-operator/aws"
+  version = "0.0.12"
+}
+
+module "airflow" {
```
BryanFauble, line 23: I deployed out airflow (It does not have access to any secrets) to the dev cluster so we could see what items are flagged in the security report

modules/sage-aws-k8s-node-autoscaler/main.tf
```diff
@@ -102,6 +102,35 @@ module "ocean-aws-k8s" {
...
+  # `desired_capacity` does not seem to force the number of nodes to increase. Look
+  # through the documentation to determine how we might manually scale up the number
+  # of nodes if we wanted to.
+  desired_capacity = var.desired_capacity
+
+
+  filters = {
+    exclude_metal = true
+    hypervisor    = ["nitro"]
```
BryanFauble, line 114: I ran into issues when I deployed airflow where the number of pods maxed out the new of private IP addresses that I could assign to the node. This should default to instances that allow trunk based ENIs, along with the `ENABLE_PREFIX_DELEGATION` in the CNI add on this should let more pods be assignable to each node.

After setting this up I could see that from https://github.com/aws/amazon-vpc-resource-controller-k8s/blob/v1.5.0/pkg/aws/vpc/limits.go - This m5.large was the same instance size I originally ran into the problem with (`m6a.large`). So it confirms that this does work.
```
	"m5.large": {
		Interface:               3,
		IPv4PerInterface:        10,
		IsTrunkingCompatible:    true,
		BranchInterface:         9,
		DefaultNetworkCardIndex: 0,
		NetworkCards: []NetworkCard{
			{
				MaximumNetworkInterfaces: 3,
				NetworkCardIndex:         0,
			},
		},
		Hypervisor:  "nitro",
		IsBareMetal: false,
	},
```

dev/spacelift/dpe-sandbox/main.tf
BWMac: A question related to this: what happens to these places where `branch` is set to `ibcdpe-1007-monitoring` when this is merged? Does it just get switched back to `main` before the merge?

dev/spacelift/dpe-sandbox/main.tf
```diff
@@ -118,15 +128,15 @@ resource "spacelift_stack_destructor" "k8s-stack-destructor" {
 
 resource "spacelift_aws_integration_attachment" "k8s-aws-integration-attachment" {
   # org-sagebase-dnt-dev-aws-integration
-  integration_id = "01J3DNYVM4AWWSDY3QEVRMQ076"
+  integration_id = "01J3R9GX6DC09QV7NV872DDYR3"
```
BWMac, line 131: Is this another thing that should be parameterized at some point (especially if we ever have both `dev` and `prod` deployments)?

modules/main.tf
```diff
@@ -1,104 +1,150 @@
...
+      repository = "eks-stack"
 
-resource "spacelift_version" "sage-aws-vpc-version" {
-  module_id      = spacelift_module.sage-aws-vpc.id
-  version_number = "0.3.3"
-}
+      name               = "sage-aws-eks"
+      terraform_provider = "aws"
+      administrative     = false
+      branch             = var.git_branch
```
BWMac, line 14: Can something like this not be used in the other places where we pass a `branch` now?

dev/spacelift/dpe-sandbox/main.tf
BryanFauble: When I merge everything back into main I will push out a change to point the branch to main

dev/spacelift/dpe-sandbox/main.tf
BryanFauble: I added these changes to https://github.com/Sage-Bionetworks-Workflows/eks-stack/pull/15

modules/main.tf
BryanFauble: I added these changes to https://github.com/Sage-Bionetworks-Workflows/eks-stack/pull/15

PR #13, opened 2024-07-23
[IBCDPE-935] VPC Updates & VPC CNI Exploration

BryanFauble:
**Problem:**

1. We needed to move the deployment over to the dnt-dev AWS account
2. We needed some more instructions on how to set up the AWS integration with spacelift and the integration with SpotIO
3. Some exploration of how to configure the VPC CNI plugin was needed.
4. Some files were no longer needed
5. A number of logs were not configured to be sent to cloud watch

**Solution:**

1. Moving to use the dnt-dev AWS account, setting up the SpotIO and spacelift integrations as well as notes on how these were set up.
2. Creating a demo module of both pod level security groups, and network policies along with recommendations within the apache airflow module about things to implement for the module.
3. Cleaning up some files that were not needed
4. Creating the configuration to export logs out to cloud watch along with a retention period of 1 day for this dev deployment.

**Testing:**

1. I have deployed everything out to the dnt-dev AWS account and that is where testing for both pod level security groups and network policies were done.

PR #15, opened 2024-08-02
[IBCDPE-1007] Templatizing items out to a root level

BryanFauble:
**Problem:**

1. As noted by @BWMac in:

* https://github.com/Sage-Bionetworks-Workflows/eks-stack/pull/14#discussion_r1700572253
* https://github.com/Sage-Bionetworks-Workflows/eks-stack/pull/14#discussion_r1700573989
* https://github.com/Sage-Bionetworks-Workflows/eks-stack/pull/14#discussion_r1700578022
* https://github.com/Sage-Bionetworks-Workflows/eks-stack/pull/13#discussion_r1700601990

1. There are a bunch of areas that are not templatized, and are using static values that we had set up for development.

**Solution:**

1. Setting up more variable and environment value passing between the stacks. Allowing us to define all of the levers within the root `main.tf` and `deployments/main.tf`
2. Removing the dev directory and instead allowing us to create different dev/prod stacks from the same terraform scripts depending on how we are defining it's usage in `deployments/main.tf`

**Testing:**

1. Tested the deployment of the dev stack and verified that it came up healthy.


deployments/spacelift/dpe-k8s/main.tf
```diff
@@ -0,0 +1,188 @@
...
+  terraform_workflow_tool = "OPEN_TOFU"
+  space_id                = spacelift_space.dpe-space.id
+}
+
+resource "spacelift_environment_variable" "k8s-stack-environment-variables" {
+  for_each = local.k8s_stack_environment_variables
+
+  stack_id   = spacelift_stack.k8s-stack.id
+  name       = "TF_VAR_${each.key}"
+  value      = try(tostring(each.value), jsonencode(each.value))
```
BryanFauble, line 62: This logic was a bit unfortunate to have to implement but it's to get around the fact that complex types being stored as environment variable as a bit wonky. See this reading:

1. https://docs.spacelift.io/vendors/terraform/storing-complex-variables
2. https://developer.hashicorp.com/terraform/language/functions/try

Without having the try and attempting to convert it to a string first I got this error:
`Invalid AWS Region: "us-east-1"`

This is because the environment variable was stored as:
`"us-east-1"`
instead of
`us-east-1`

modules/trivy-operator/main.tf
```diff
@@ -17,27 +17,22 @@ resource "helm_release" "trivy-operator" {
...
-        }
-      ]
-      selector = {
-        matchLabels = {
-          "app.kubernetes.io/name" = "trivy-operator"
-        }
-      }
-    }
-  }
+resource "kubectl_manifest" "vmservicescrape" {
```
BryanFauble, line 20: Swapping over to this is to resolve a race condition issue. Since the VMServiceScrape kind is a resource that depends on a CRD (Custom Resource Definition) being installed already on the cluster the terraform plan stage was failing with:

```
│ Error: API did not recognize GroupVersionKind from manifest (CRD may not be installed)
│ 
│   with module.trivy-operator.kubernetes_manifest.vmservicescrape,
│   on .terraform/modules/trivy-operator/main.tf line 20, in resource "kubernetes_manifest" "vmservicescrape":
│   20: resource "kubernetes_manifest" "vmservicescrape" {
│ 
│ no matches for kind "VMServiceScrape" in group
│ "operator.victoriametrics.com"
```

The suggestion was to try out a different provider (https://registry.terraform.io/providers/gavinbunney/kubectl/latest/docs/resources/kubectl_manifest) to get around this issue. This allows the resources to be created as expected in the right order.

BWMac: Interesting that thisre resource has you pass a whole YAML configuration to the `yaml_body` arg instead of somthing more similar to how `kubernetes_manifest` is configured

modules/victoria-metrics/templates/values.yaml
```diff
@@ -16,7 +16,7 @@ victoria-metrics-operator:
   cleanupImage:
     repository: bitnami/kubectl
     # use image tag that matches k8s API version by default
-    # tag: 1.29.6
+    tag: 1.30.2
```
BWMac, line 19: Is pinning this version better than using the default as was done before?

deployments/airflow/Dockerfile
```diff
@@ -1,13 +0,0 @@
-ARG BASE_IMAGE=apache/airflow:2.7.1-python3.10
```
BWMac, line 1: With the airflow `deployment` directory being removed, how does the deployment of airflow with our configurations work now?

BryanFauble: These files were never used to begin with.

modules/victoria-metrics/templates/values.yaml
BryanFauble: There was a bug that I had reported around this. Basically a '+' sign that was getting included in the tag image that prevented the cleanup job from running. I updated the helm chart to the newer version so we don't need to set this tag manually anymore. I will remove

BWMac (approved):

PR #17, opened 2024-08-15
[IBCDPE-1005] eks developer role

BryanFauble:
**Problem:**

1. As a user with a Developer level role I do not have access to the EKS cluster in any way.

**Solution:**

1. Creating an access entry for both Administrator and Developer SSO roles that have an EKS policy to access resources on the cluster
2. This gives cluster level read access: https://docs.aws.amazon.com/eks/latest/userguide/access-policies.html

**Testing:**

1. I updated my kube config and verified I could access resources from both sso roles
3. I tested that I was able to view all resources and perform port-forwarding sessions
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,11 +135,11 @@ aws sso login --profile dpe-prod-admin
 # AWS using my SSO session for the profile `dpe-prod-admin`. After authenticated 
 # assuming that we want to use the `role/eks_admin_role` to connect to the k8s 
 # cluster". This will update your kubeconfig with permissions to access the cluster.
-aws eks update-kubeconfig --region us-east-1 --name dpe-k8 --role-arn arn:aws:iam::766808016710:role/eks_admin_role --profile dpe-prod-admin
+aws eks update-kubeconfig --region us-east-1 --name dpe-k8 --profile dpe-prod-admin
 ```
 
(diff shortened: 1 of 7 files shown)


thomasyu888 (approved):
🔥 LGTM!  Going to pre-approve - not sure if @BWMac had some comments!

BWMac (approved):

zaro0508 (changes requested):
i think this is the wrong approach, see my comments in PR https://github.com/Sage-Bionetworks-IT/organizations-infra/pull/1191

BryanFauble:
> i think this is the wrong approach, see my comments in PR [Sage-Bionetworks-IT/organizations-infra#1191](https://github.com/Sage-Bionetworks-IT/organizations-infra/pull/1191)

@zaro0508 You bring up an excellent point that I didn't think of until you brought this up.

I actually don't need to create a new role for one to assume in this case. What I can do instead is to add these existing assumed roles as an IAM access entry to the cluster. This prevent the assume for any user to assume another role with these permissions, and instead I can directly grant these permissions like:
image


I will update this PR accordingly

thomasyu888 (approved):
Great changes and thanks for the review khai!

PR #22, opened 2024-08-17
[IBCDPE-1004] Implement Postgres server

BryanFauble:
TODO: Clean up the branch before marking the PR for ready for review. I had deployed out a testing stack in spacelift.

**Problem:**

1. Airflow did not recommend using the airflow instance from their helm chart for a production load due to the maintenance requirements of handling it.

**Solution:**

1. Implementing https://cloudnative-pg.io/documentation/current/
2. Updating the connections for airflow to use the new connection and set up the account secrets automatically.

**Description of tool:**

> CloudNativePG is an open source [operator](https://kubernetes.io/docs/concepts/extend-kubernetes/operator/) designed to manage [PostgreSQL](https://www.postgresql.org/) workloads on any supported [Kubernetes](https://kubernetes.io/) cluster. It defines a new Kubernetes resource called Cluster representing a PostgreSQL cluster made up of a single primary and an optional number of replicas that co-exist in a chosen Kubernetes namespace for High Availability and offloading of read-only queries.

> Applications that reside in the same Kubernetes cluster can access the PostgreSQL database using a service which is solely managed by the operator, without having to worry about changes of the primary role following a failover or a switchover.

**Testing:**

1. Verified that a deployment of the operator and cluster for airflow worked, airflow came up healthy. I was able to log in, and trigger a DAG.


BryanFauble:
Will open this back up shortly

PR #23, opened 2024-08-19
[IBCDPE-1004] Airflow Prod + Postgresql-cloud native

BryanFauble:
**Problem:**

1. A production postgres instance was needed for our airflow deployment
2. Initial airflow secrets were manually hard coded in

**Solution:**

1. Setting up https://cloudnative-pg.io/ and creating an application database for postgres
2. Set up a random admin secret for the first airflow user
3. Adding a grafana dashboard for postgres

**Testing:**

1. I deployed this out to our dev cluster


main.tf
```diff
@@ -10,7 +10,7 @@
 # }
 
 locals {
-  git_branch = "main"
+  git_branch = "ibcdpe-1004-airflow-ops"
```
BryanFauble, line 13: To swap back after merge.

My plan in the future is to allow us to stand up the entire infra stack when we want to test items. I had some issues with tearing down the VPC which prevented me from wanting to do this. Once those are ironed out we'll be able to start up independent infra for testing, then tear it down when we are done.

modules/victoria-metrics/templates/values.yaml
```diff
@@ -1143,7 +1147,7 @@ crds:
 
 ## install prometheus operator crds
 prometheus-operator-crds:
-  enabled: false
+  enabled: true
```
BryanFauble, line 1150: By installing these CRDs we can install prometheus specific resources and victoria metrics will create the relevant victoria metric resource. This is useful when installing helm charts that have defined scrapes and other prometheus configs.

deployments/stacks/dpe-k8s-deployments/main.tf
```diff
@@ -31,14 +31,34 @@ module "trivy-operator" {
 module "airflow" {
   depends_on   = [module.victoria-metrics, module.argo-cd, module.sage-aws-eks-autoscaler]
   source       = "spacelift.io/sagebionetworks/airflow/aws"
-  version      = "0.3.1"
+  version      = "0.3.5"
```
thomasyu888, line 34: Are we self defining the version of these?

modules/postgres-cloud-native/templates/cluster-values.yaml
thomasyu888, line 1: Anything special about this configuration?

thomasyu888 (approved):
🔥 Going to pre-approve, as this looks good.  Theres a lot of configurations that we are adding, so I was just triple checking all the different password configurations.

Neat that you kubernetes can auto generate random passwords and store it as a data point.

modules/main.tf
```diff
@@ -100,7 +100,7 @@ locals {
       description        = "Helm chart deployment for apache airflow."
       project_root       = "modules/apache-airflow"
       space_id           = "root"
-      version_number     = "0.3.1"
+      version_number     = "0.3.5"
```
BWMac, line 103: I guess another candidate for parameterization could be the places where our version numbers are. If we had one place which holds all of the version numbers and then all of the references could then be pointed at that param, then you only need to increment it in one place.

modules/postgres-cloud-native-operator/README.md
```diff
@@ -0,0 +1,24 @@
+# Purpose
+The purpose of this module is to deploy the `Cloudnative PG` helm chart <https://github.com/cloudnative-pg/charts/tree/main>.
+This will deploy both the operator.
```
BWMac, line 3: Is this sentence missing something?

BWMac (approved):

deployments/stacks/dpe-k8s-deployments/main.tf
BryanFauble: Yes. When we increment a module in `/modules/main.tf` it pushes out a new version to the spacelift terraform registry:
image

We can then install any version of these as well as make breaking changes without affecting anything that might be using older versions. It is slightly overkill since everything is in a single repo (And you could reference it via relative importing). But if we start by following this it should scale out to handle for cases where we have many different deployments using different versions, potentially spread across multiple repos.

modules/postgres-cloud-native/templates/cluster-values.yaml
BryanFauble: Some things to point out:

We're defaulting storage to 8Gi, but volume expansion is turned on. So when we are getting close we can bump it up and swap it to be configurable:
```
  storage:
    size: 8Gi
    storageClass: ""
```

Monitoring/pod monitoring is enabled. This resource definition for promethues CRDs which is then intercepted by victoria metrics to create their specific formatted resource definition. It's is what enables metric scraping for viewing in grafana:
```
  monitoring:
    # -- Whether to enable monitoring
    enabled: true
    podMonitor:
      # -- Whether to enable the PodMonitor
      enabled: true
```

The database that gets created will be called `application-database` and the initial application user is based off the secret created in the `main.tf` file:
```
  initdb:
    database: application-database
    owner: "" # Defaults to the database name
    secret: 
      name: "pg-user-secret" # Name of the secret containing the initial credentials for the owner of the user database. If empty a new secret will be created from scratch
```

Backups are a future enhancement to offload the data to S3

modules/main.tf
BryanFauble: I'm not sure if I am totally on-board extracting it out to a single location since we might be in situations we want to test version A in a stack, but continue to run version B in another stack.

But I agree, there is work to improve this.

BryanFauble:
> Neat that you kubernetes can auto generate random passwords and store it as a data point.

Kind of. In this case terraform is creating a random password. That is stored in the terraform state. That random password is then passed to a kubernetes secret which is then used as an environment variable in the pods it's needed.

There is some future stuff I want to eventually explore something like:

1. https://github.com/bitnami-labs/sealed-secrets
2. Or more extensive use of AWS secret manager

PR #25, opened 2024-08-20
[IBCDPE-1059] Upgrade autoscaler

BryanFauble:
**Problem:**

1. V1 of the autoscaler will be end of life in november

**Solution:**

1. Upgrade per their documentation: https://docs.spot.io/ocean/tutorials/spot-kubernetes-controller/ocean-controller-two-install?id=install-via-terraform

**Testing:**

1. I tested the deployment and promoted the changes. Diff shows no changes now:
image

**Additional changes:**

Enable the cleanup job for airflow, otherwise even if the task fails in airflow the pod looks like it sticks around:

image
image


modules/apache-airflow/templates/values.yaml
```diff
@@ -2037,13 +2037,13 @@ cleanup:
...
-  #  requests:
-  #   cpu: 100m
-  #   memory: 128Mi
+  resources:
+   limits:
+    cpu: 100m
+    memory: 128Mi
+   requests:
+    cpu: 100m
+    memory: 128Mi
```
BWMac, line 2046: These resource allocations seem pretty minimal to me. Is that accurate, or am I off base there? Is there any reason to bump it up?

BWMac (approved):
Looks good, just one question

modules/apache-airflow/templates/values.yaml
BryanFauble: This was their recommended defaults. We can adjust these later on by looking at historical data in grafana.

PR #27, opened 2024-08-20
[IBCDPE-1004] Create connection pooler

BryanFauble:
Problem:

1. Looking at database stats Apache airflow was taking all 100 connections and not giving them up:
image

From docs I read this is a known item and their own recommendation is to use pgbouncer as a connection pooler.

Solution:

1. The cloud native PG instance has an easy way to implement a connection pooler. Turning it on and pointing the connection secret to the connection pooler.

Testing:

1. Will deploy and test on the sandbox cluster.

### DIFF
--- a/modules/postgres-cloud-native/main.tf
+++ b/modules/postgres-cloud-native/main.tf
@@ -34,7 +34,6 @@ spec:
     kustomize:
       patches:
       - target:
-          kind: Pooler
           name: ${var.argo_deployment_name}-cluster
         patch: |-
             - op: replace
--- a/modules/postgres-cloud-native/main.tf
+++ b/modules/postgres-cloud-native/main.tf
@@ -29,7 +29,7 @@ spec:
     targetRevision: ${var.git_revision}
     ref: values
   - repoURL: 'https://github.com/Sage-Bionetworks-Workflows/eks-stack.git'
-    targetRevision: ibcdpe-1004-add-pooler
+    targetRevision: ${var.git_revision}
     path: modules/postgres-cloud-native/resources
     kustomize:
       patches:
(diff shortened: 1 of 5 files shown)


modules/postgres-cloud-native/main.tf
```diff
@@ -54,14 +71,14 @@ resource "kubernetes_secret" "connection-secret" {
...
     "password"   = random_password.pg-password.result
-    "pgpass"     = "${var.argo_deployment_name}-cluster-rw:5432:application-database:application-database:${random_password.pg-password.result}"
+    "pgpass"     = "${var.argo_deployment_name}-pooler-rw:5432:application-database:application-database:${random_password.pg-password.result}"
     "port"       = "5432"
-    "uri"        = "postgresql://application-database:${random_password.pg-password.result}@${var.argo_deployment_name}-cluster-rw.${var.namespace}:5432/application-database"
+    "uri"        = "postgresql://application-database:${random_password.pg-password.result}@${var.argo_deployment_name}-pooler-rw.${var.namespace}:5432/application-database"
     "user"       = "application-database"
     "username"   = "application-database"
-    "connection" = "postgresql://application-database:${random_password.pg-password.result}@${var.argo_deployment_name}-cluster-rw.${var.namespace}:5432/application-database"
+    "connection" = "postgresql://application-database:${random_password.pg-password.result}@${var.argo_deployment_name}-pooler-rw.${var.namespace}:5432/application-database"
```
BryanFauble, line 82: A future enhancement would be to allow the person using this module to create a connection pooler or not. Likely they would have a sensible application that doesn't take every connection to the DB possible.

BWMac (approved):

PR #29, opened 2024-08-21
[IBCDPE-1004] Set env for class deserialization

BryanFauble:
**Problem:**

1. `ImportError: unusual_prefix_b22621ef2028bf7db4581afeff23e3f182d62fc3_trending-projects-snapshot-dag.SnapshotMetrics was not found in allow list for deserialization imports. To allow it, add it to allowed_deserialization_classes in the configuration`
2. All of the classes that are passed around are not explicitly allowed, running into this error.

**Solution:**

1. Adding an environment variable that is supposed to allow this to allow anything so we don't need to go in and re-deploy airflow.

**Testing:**

1. Verified after deploying changes that the content is making it through by testing this dag: https://github.com/Sage-Bionetworks-Workflows/orca-recipes/blob/test-deserialization/dags/testing-deserialization.py

image

### DIFF
--- a/modules/apache-airflow/templates/values.yaml
+++ b/modules/apache-airflow/templates/values.yaml
@@ -1055,14 +1055,14 @@ webserver:
   #      - "foo.remote"
   allowPodLogReading: true
   livenessProbe:
-    initialDelaySeconds: 60
+    initialDelaySeconds: 120
     timeoutSeconds: 5
     failureThreshold: 5
     periodSeconds: 10
     scheme: HTTP
 
   readinessProbe:
-    initialDelaySeconds: 60
+    initialDelaySeconds: 120
     timeoutSeconds: 5
     failureThreshold: 5
     periodSeconds: 10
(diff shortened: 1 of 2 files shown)


modules/apache-airflow/templates/values.yaml
```diff
@@ -2105,6 +2103,7 @@ config:
     # For Airflow 1.10, backward compatibility; moved to [logging] in 2.0
     colored_console_log: 'False'
     remote_logging: '{{- ternary "True" "False" .Values.elasticsearch.enabled }}'
+    allowed_deserialization_classes: "airflow.* DownloadMetric SnapshotMetrics LENSDataset Dataset Syn.*"
```
thomasyu888, line 2106: Nice work here, is this a risk if we have to keep adding to this list?

BryanFauble: I think having to specify this is a ridiculous requirement that they added. I am trying to see if I can just wild card allow anything. Then if it cannot be de serialized, then fail at runtime.

thomasyu888 (approved):
🔥LGTM!

PR #30, opened 2024-08-22
[IBCDPE-1004] Pull back to single AZ

BryanFauble:
**Problem:**

1. EBS volumes being used for mounted storage within the EKS cluster can only be re-mounted in the same AZ that it was created in. This is a problem as Spot IO when it provisions new nodes for the clusters does not use this context to know when to provision another cluster in the appropriate AZ. This leads to pods being stuck in pending and not scheduled.

**Solution:**

1. Fall back to a single AZ for now. Later on we'll determine a more robust solution to span AZs for High availability.


thomasyu888 (approved):
🔥 LGTM!

PR #34, opened 2024-09-12
Updating permissions for eks developer to allow edit

BryanFauble:
**Problem:**

1. The admin view role does not permit port-forwarding.

**Solution:**

1. Allowing a non-admin use of edit that includes permissions to start a port-forwarding session


thomasyu888 (approved):
🔥 LGTM!

PR #35, opened 2024-10-01
[IBCDPE-938] Deploy Signoz (OTEL Visualization) to kubernetes cluster

BryanFauble:
Document describing the overall design of the solution and some included network diagrams at the bottom:
https://sagebionetworks.jira.com/wiki/spaces/DPE/pages/3351773274/IBCDPE-938+OpenTelemetry+Data+collection+Visualization

**Problem:**

1. We did not have a way to view telemetry data produced by services like schematic
2. We were deploying EC2 instances to a single AZ, public subnets were not deployed to all AZs.
3. We did not have a north-south gateway to handle traffic into the cluster (From internet into cluster)
4. We did not have AWS SES to send out emails
5. We did not have any TLS set up for secure messaging into the cluster
6. We did not have authentication mechanisms set up to prevent unauthenticated traffic from getting ingested into the cluster

**Solution:**

1. Deploy the signoz helm chart: https://github.com/SigNoz/charts/tree/main/charts/signoz - Which include the clickhouse operator for data storage
2. Availability zone: Updating AZ deployment to deploy worker nodes on 3 private subnets, create 3 public subnets, and deploy EKS control plane on 2 AZs
7. envoy-gateway is added to the cluster to handle north-south traffic: https://gateway.envoyproxy.io/docs/ - This also gives us an easy way to secure the endpoints with auth filters/expose future services to the internet.
8. cert-manager is added to the cluster to handle TLS/cert generation
9. AWS SES module created to configure SES settings to allow SigNoz alert manager component to send out emails.
10. Setting up Auth0 to allow clients to request an access token from Auth0 that can be verified at the gateway within the k8s cluster.

**Testing:**

1. I have tested the deployment of signoz to the sandbox cluster in the DNT dev account
2. I have tested the north-south traffic provided through envoy-gateway
3. I have tested with auth0 that JWT verification works through the envoy-gateway
4. I tested sending the data to a backend service in the k8s cluster to verify traffic makes it to the pod
5. I tested and provisioned a cert from let's encrypt with cert-manager for HTTPS/TLS
12. I tested that pods are started in on-demand instances when they have the nodeSelector set
13. I tested AWS SES and have successfully requested production level access for the SES account in our sandbox cluster

Untitled

**Follow up tickets:**

- [ ]  Set up Clickhouse Cold Storage + Backups (Does not block this review): https://sagebionetworks.jira.com/browse/IBCDPE-1094
- [X] Create cluster ingress for otel data: https://sagebionetworks.jira.com/browse/IBCDPE-1095
- [X] Depending on response from Create sub-domain for DPE Kubernetes cluster ingress: https://sagebionetworks.jira.com/browse/IT-3931 - We may have a subdomain to use for all DNS records and some additional work may be needed to pull TLS certs in a centralized manner as defined by IT. This may also influence some decisions on handling authentication between schematic and the k8s cluster.


deployments/stacks/dpe-k8s-deployments/main.tf
```diff
@@ -21,8 +21,9 @@ module "sage-aws-eks-addons" {
 
 module "argo-cd" {
   depends_on = [module.sage-aws-eks-autoscaler]
-  source     = "spacelift.io/sagebionetworks/argo-cd/aws"
-  version    = "0.3.1"
+  # source     = "spacelift.io/sagebionetworks/argo-cd/aws"
+  # version    = "0.3.1"
+  source = "../../../modules/argo-cd"
 }
```
BryanFauble, line 27: I am liking the usage of this relative source import, especially for testing.

deployments/main.tf
```diff
@@ -37,15 +37,7 @@ module "dpe-sandbox-spacelift-development" {
...
-  # azs                  = ["us-east-1a", "us-east-1b", "us-east-1c"]
-  # For now, we are only using one public and one private subnet. This is due to how
-  # EBS can only be mounted to a single AZ. We will need to revisit this if we want to
-  # allow usage of EFS ($$$$), or add some kind of EBS volume replication.
-  # Note: EKS requires at least two subnets in different AZs. However, we are only using
-  # a single subnet for node deployment.
+  vpc_cidr_block       = "10.51.0.0/16"
   public_subnet_cidrs  = ["10.51.1.0/24", "10.51.2.0/24"]
   private_subnet_cidrs = ["10.51.4.0/24", "10.51.5.0/24"]
   azs                  = ["us-east-1a", "us-east-1b"]
```
BryanFauble, line 43: I am leaving this on a 2 AZ deployment (Since it cannot be changed after the k8s cluster is created). I did set the single AZ part of `deployments/stacks/dpe-k8s-deployments/main.tf` to false so that the spot io cluster autoscaler can deploy EC2 instances to either private subnet

modules/apache-airflow/templates/values.yaml
```diff
@@ -108,7 +108,9 @@ images:
     pullPolicy: IfNotPresent
 
 # Select certain nodes for airflow pods.
-nodeSelector: {}
+nodeSelector: {
+  failure-domain.beta.kubernetes.io/zone: us-east-1a
+}
```
BryanFauble, line 113: Note the changes in this values.yaml file to update replicas from 2 to 1. Running these services with on-demand instances will get us towards the stability we wanted. Running these workloads on spot instances was not the right decision.

modules/apache-airflow/templates/values.yaml
```diff
@@ -636,7 +638,9 @@ workers:
   extraVolumeMounts: []
 
   # Select certain nodes for airflow worker pods.
-  nodeSelector: {}
+  nodeSelector: {
+    spotinst.io/node-lifecycle: "od"
+  }
```
BryanFauble, line 643: This node selector defined how the spot io cluster autoscaler requests EC2 instances to spin up for the cluster. This makes sure that an on-demand instance is created and that the pods are running on it.

modules/argo-cd/templates/values.yaml
```diff
@@ -501,7 +501,14 @@ configs:
...
   ## Note: the last example in the list would use a repository credential template, configured under "configs.credentialTemplates".
-  repositories: {}
+  repositories:
+    docker-registry:
+      url: registry-1.docker.io
+      # username: "docker"
+      # password: ""
+      name: docker-registry
+      enableOCI: "true"
+      type: "helm"
```
BryanFauble, line 511: When I was testing some helm charts that use the `oci://` prefix, this was needed.

modules/cluster-ingress/README.md
```diff
@@ -0,0 +1,49 @@
+# Purpose
+The purpose of this module is to deploy kubernetes resources related to ingress for
+the cluster. Along with the ingress we will also deploy out the related SSL cert issuer.
+
+## To implemement
+The Envoy Gateway can secure ingress by verifying JWT. It can be applied to a specific
+target, for example this applies it to an HTTPRoute called `backend`:
```
BryanFauble, line 7:
```suggestion
target, for example this applies it to all requests going through a `Gateway` called `eg`
```

modules/envoy-gateway/create-synapse-oauth-client.py
```diff
@@ -0,0 +1,26 @@
+import synapseclient
+import json
+syn = synapseclient.login()
```
BryanFauble, line 3: We might be able to create a client for each service that communicates with the cluster. Thus offloading the responsibility to get an auth token on the caller. Then - in the envoy gateway, we have a JWT auth provider with 1 or more providers to verify the signature.

modules/cluster-ingress/README.md
```diff
@@ -0,0 +1,49 @@
...
+  targetRef:
+    group: gateway.networking.k8s.io
+    kind: Gateway
+    name: eg
+  jwt:
+    providers:
+    - name: auth0
+      remoteJWKS:
+        uri: https://dev-57n3awu5je6q653y.us.auth0.com/.well-known/jwks.json
+```
```
BryanFauble, line 24: I thought this was a really nifty way to define how the envoy gateway is able to authenticate requests passing through it. There are a number of pre-defined ways shown here: https://gateway.envoyproxy.io/latest/tasks/security/jwt-authentication/ that will allow us to integrate with auth providers.

Another example here for the OIDC Authentication shows how we might be able to do something like stand up a user login screen in front of the cluster using Synapse (Such as how DCA works):
https://gateway.envoyproxy.io/latest/tasks/security/oidc/

modules/cluster-ingress/main.tf
```diff
@@ -0,0 +1,40 @@
...
+      patches:
+      - target:
+          kind: Gateway
+        patch: |-
+          - op: replace
+            path: /metadata/annotations/cert-manager.io~1cluster-issuer
+            value: ${var.cluster_issuer_name}
+          - op: replace
+            path: /spec/listeners/0/hostname
+            value: ${var.ssl_hostname}
```
BryanFauble, line 29: This ssl_hostname is a little rough in it's current form. In order to know what the hostname is going to be this module has to be deployed, and we need to get the DNS name of the AWS load balancer that is created, then apply the change again with the correct value.

Ideally, if we get a subdomain to use that is already created we will have this static value ahead of time to pass in here.

modules/apache-airflow/templates/values.yaml
```diff
@@ -806,7 +810,9 @@ scheduler:
   extraVolumeMounts: []
 
   # Select certain nodes for airflow scheduler pods.
-  nodeSelector: {}
+  nodeSelector: {
+    spotinst.io/node-lifecycle: "od"
```
thomasyu888, line 814: Maybe for later we can think about other spot setup configuraitons that would make Airflow more resilient so we actually gain the benefit of the spot instances.

modules/cluster-ingress/README.md
thomasyu888: That's really neat!  I wonder how feasible that would be.

modules/cluster-ingress/resources/gateway.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: gateway.networking.k8s.io/v1
+kind: Gateway
+metadata:
+  name: eg
+  annotations:
+    cert-manager.io/cluster-issuer: <cluster-issuer-placeholder>
```
thomasyu888, line 6: Is this where we have to add the value back in after the resource is deployed?

modules/envoy-gateway/create-synapse-oauth-client.py
thomasyu888: Oh interesting - for every service, we could auto-create an oauth client? Not sure if we've tested this out explicitly with envoy gateway, or if thats another ticket.

modules/signoz/README.md
```diff
@@ -0,0 +1,62 @@
...
+Local Port:     4317,4318
+```
+
+Mac:
+```
+Container Port: collector::4317,collector::4318
+Local Port:     4317,4318
+```
+
+Some data will be present in those fields by default, delete was is there and copy the
```
thomasyu888, line 55:
```suggestion
Some data will be present in those fields by default, delete what is there and copy the
```

modules/signoz/templates/values.yaml
```diff
@@ -0,0 +1,2393 @@
...
+  database: signoz_metrics
+  # -- Clickhouse trace database (SigNoz Traces)
+  traceDatabase: signoz_traces
+  # -- Clickhouse log database (SigNoz Logs)
+  logDatabase: signoz_logs
+  # -- Clickhouse user
+  user: admin
+  # -- Clickhouse password
+  # TODO: Replace with random password
+  password: 27ff0399-0d3a-4bd8-919d-17c2181e6fb9
```
thomasyu888, line 131: Is this an issue that this is committed right now?

modules/signoz/templates/values.yaml
thomasyu888, line 1: Is most of this template default values?

thomasyu888 (approved):
🔥 LGTM! Could you provide me with port-forward instructions to the signoz service?  There seems to be spacelift conflicts in deployment, do all the argo deployments need to be re-versioned?

deployments/main.tf
BryanFauble: No, This should work in conjunction with this change to limit the deployment to nodes in a specific zone, while still allowing nodes to spin up in either zone:

```
nodeSelector: {
  failure-domain.beta.kubernetes.io/zone: us-east-1a
}
```

modules/envoy-gateway/create-synapse-oauth-client.py
BryanFauble: For our use-case with machine-to-machine authentication or the `client credentials grant type from OAuth` we cannot use Synapse as it is not supported.

What we can use is Auth0 for this purpose. I did a small PoC that I could generate and auth an Auth0 token, but I hadn't set up the full round-trip for it, I also want to test out the extension part of the otel exporter as this is what Schematic would need to do on their side: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/oauth2clientauthextension

modules/cert-manager/templates/values.yaml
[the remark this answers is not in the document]
BryanFauble: I did not need to make any changes here, and everything looked like good defaults.

modules/apache-airflow/templates/values.yaml
BryanFauble: For airflow I think we can take advantage of this if we move over to kubernetes workers instead of celery works. The reason is that the individual tasks can run on spot instances, but everything to kick off those tasks don't do well to be interrupted on spot instances.

modules/cluster-ingress/resources/gateway.yaml
BryanFauble: This is handled in the `modules/cluster-ingress/main.tf` file:

```
    kustomize:
      patches:
      - target:
          kind: Gateway
        patch: |-
          - op: replace
            path: /metadata/annotations/cert-manager.io~1cluster-issuer
            value: ${var.cluster_issuer_name}
```

modules/signoz/templates/values.yaml
BryanFauble: I set a few sections of:
```
  nodeSelector: {
    spotinst.io/node-lifecycle: "od"
  }
```

I set this to have 1 shard and 2 replicas:
https://clickhouse.com/docs/en/architecture/replication


Cold storage has not been setup yet (There is a ticket for it).


I updated the otel collector to have 2 replicas instead of 1.


The rest should be default values.

BryanFauble:
> 🔥 LGTM! Could you provide me with port-forward instructions to the signoz service? There seems to be spacelift conflicts in deployment, do all the argo deployments need to be re-versioned?

Port-forwading:
https://emmer.dev/blog/port-forwarding-to-kubernetes/#k9s

You can either use the service called `signoz-frontend` or the pod `signoz-frontend-id_values`.

When you land at the initial login page I'll need to send you a token, reach out to me in slack and we will coordinate that.

deployments/main.tf
BryanFauble: I also created https://sagebionetworks.jira.com/browse/IBCDPE-1097 as a follow-up here: Move worker node subnet off EKS cluster subnets

From these docs:  https://aws.github.io/aws-eks-best-practices/networking/subnets/#vpc-configurations


> Kubernetes worker nodes can run in the cluster subnets, but it is not recommended. During [cluster upgrades](https://aws.github.io/aws-eks-best-practices/upgrades/#verify-available-ip-addresses) Amazon EKS provisions additional ENIs in the cluster subnets. When your cluster scales out, worker nodes and pods may consume the available IPs in the cluster subnet. Hence in order to make sure there are enough available IPs you might want to consider using dedicated cluster subnets with /28 netmask.

modules/envoy-gateway/create-synapse-oauth-client.py
```diff
@@ -0,0 +1,26 @@
+import synapseclient
```
zaro0508, line 1: I'm wondering how this file is triggered or used?

BryanFauble: It's not at the moment. I originally had the thought that we might be able to use Synapse for the Oauth client credential flow (I learned we cannot), however, we will eventually be able to use the OpenID Connect (OIDC) flow for securing web browser access.

modules/signoz/README.md
```diff
@@ -0,0 +1,62 @@
+# Purpose
+The purpose of this module is to deploy the `Signoz` helm chart <https://github.com/SigNoz/charts/tree/main/charts/signoz>.
+
+SigNoz is an open-source APM. It helps developers monitor their applications 
+& troubleshoot problems, an open-source alternative to DataDog, NewRelic, etc. Open 
+source Application Performance Monitoring (APM) & Observability tool.
+
+
+## This module is a work in progress
```
zaro0508, line 9: which is work in progress? the [signoz terraform module](https://registry.terraform.io/providers/SigNoz/signoz/latest/docs) itself?  or this specific setup of signoz?

modules/signoz/README.md
```diff
@@ -0,0 +1,62 @@
...
+A number of items are needed:
+
+- Setting up backups and data retention: https://sagebionetworks.jira.com/browse/IBCDPE-1094
+- Set up ingress to the cluster/collector to send data to: https://sagebionetworks.jira.com/browse/IBCDPE-1095
+- Set up accounts and access to the service decleratively
+
+## Accessing signoz
```
zaro0508, line 16: SageIT would be very interested to have this setup with SSO+SCIM to help with onboarding/offboard users, otherwise unauthorized user access will be inevitable.

BryanFauble: My thought was to use the Synapse OIDC flow for users accessing the UI. I need to further document me exact plan but the idea was:

1. Create a client to authenticate users with Synapse
2. Verify the users membership to a specific Synapse group
3. As users are added/removed from the Synapse group they will gain/lose access to the UI

Signoz self-hosted doesn't integrate with SSO. So what we would have to do is have users log in through Synapse oidc (Which IT could help control), and then further log in through the Signoz user interface to their specific account.

Thoughts?

modules/signoz/README.md
BryanFauble: I noted the item below.

I am not using the terraform provider that you linked to, it does nit implement anything but a declarative configuration for some alerts.

The tool (spacelift) that I am using for Deploying the terraform scripts has a private module registry. That documentation is here:
https://docs.spacelift.io/vendors/terraform/module-registry


That private module registry would let us tag specific module versions that we create (Everything in the /modules directory of this project). It also lets us easily register modules from other git repositories, meaning we are not restricted to a mono repo style of code, we can use a mix.

modules/envoy-gateway/create-synapse-oauth-client.py
BryanFauble: I will now that I have a better idea how I'll use this. See this comment chain:
https://github.com/Sage-Bionetworks-Workflows/eks-stack/pull/35#discussion_r1797202156

deployments/main.tf
```diff
@@ -37,15 +37,7 @@ module "dpe-sandbox-spacelift-development" {
...
-  vpc_cidr_block = "10.51.0.0/16"
-  # public_subnet_cidrs  = ["10.51.1.0/24", "10.51.2.0/24", "10.51.3.0/24"]
-  # private_subnet_cidrs = ["10.51.4.0/24", "10.51.5.0/24", "10.51.6.0/24"]
-  # azs                  = ["us-east-1a", "us-east-1b", "us-east-1c"]
-  # For now, we are only using one public and one private subnet. This is due to how
-  # EBS can only be mounted to a single AZ. We will need to revisit this if we want to
-  # allow usage of EFS ($$$$), or add some kind of EBS volume replication.
-  # Note: EKS requires at least two subnets in different AZs. However, we are only using
-  # a single subnet for node deployment.
+  vpc_cidr_block       = "10.51.0.0/16"
```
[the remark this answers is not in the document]
BryanFauble: That should work for my purpose. I will look into making this change.

modules/cluster-ingress/main.tf
BryanFauble: @zaro0508 i have this ticket out there:

https://sagebionetworks.jira.com/browse/IT-3931

My thought was to use an existing subdomain, however, i am also open to an entirely new domain for this as well. Im fine with whatever suggestion that you have for this.

deployments/main.tf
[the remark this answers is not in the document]
BryanFauble: I made these changes and put them into another PR:
https://github.com/Sage-Bionetworks-Workflows/eks-stack/pull/39

README.md
```diff
@@ -283,10 +291,27 @@ This document describes the abbreviated process below:
...
+                "iam:ListPolicies",
+                "iam:TagUser",
+                "iam:GetUser",
+                "iam:DeleteUser",
+                "iam:CreateAccessKey",
+                "iam:ListAccessKeys",
+                "iam:DeleteAccessKeys"
+            ],
+            "Resource": "arn:aws:iam::{{AWS ACCOUNT ID}}:user/smtp_user"
+        }
```
BryanFauble, line 314: Required for the IaC to set up the SMTP user. Is this acceptable, or should this user be created manually?

deployments/main.tf
```diff
@@ -31,24 +31,49 @@ module "dpe-sandbox-spacelift-development" {
...
+      description = "Client for Github Actions to export telemetry data"
+      app_type    = "non_interactive"
+    },
+    {
+      name        = "schematic - Dev"
+      description = "Client for schematic deployed to AWS DEV to export telemetry data"
+      app_type    = "non_interactive"
+    },
+  ]
+
```
BryanFauble, line 54: This is how we'll set up new account to get access to send telemetry data and authenticate with Auth0. It also relates to some of the documentation I wrote here: https://github.com/Sage-Bionetworks/schematic/pull/1527/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5

deployments/main.tf
```diff
@@ -31,24 +31,49 @@ module "dpe-sandbox-spacelift-development" {
...
-  private_subnet_cidrs = ["10.51.4.0/24", "10.51.5.0/24"]
-  azs                  = ["us-east-1a", "us-east-1b"]
+  vpc_cidr_block = "10.52.16.0/20"
+  # A public subnet is required for each AZ in which the worker nodes are deployed
+  public_subnet_cidrs                    = ["10.52.16.0/24", "10.52.17.0/24", "10.52.19.0/24"]
+  private_subnet_cidrs_eks_control_plane = ["10.52.18.0/28", "10.52.18.16/28"]
+  azs_eks_control_plane                  = ["us-east-1a", "us-east-1b"]
+
+  private_subnet_cidrs_eks_worker_nodes = ["10.52.28.0/22", "10.52.24.0/22", "10.52.20.0/22"]
+  azs_eks_worker_nodes                  = ["us-east-1c", "us-east-1b", "us-east-1a"]
```
BryanFauble, line 68: This shrinks the VPC size. The order of the AZs also means:

1. EKS Control plane is deployed to 2 AZs
2. Public subnets are deployed to all 3 AZs
3. Private subnets are deployed to all 3 AZs

deployments/main.tf
```diff
@@ -31,24 +31,49 @@ module "dpe-sandbox-spacelift-development" {
...
+
+  enable_cluster_ingress = true
+  enable_otel_ingress    = true
+  ssl_hostname           = "dev.sagedpe.org"
+  auth0_jwks_uri         = "https://dev-sage-dpe.us.auth0.com/.well-known/jwks.json"
+  ses_email_identities   = ["aws-dpe-dev@sagebase.org"]
+  # Defines the email address that will be used as the sender of the email alerts
+  smtp_from = "aws-dpe-dev@sagebase.org"
```
BryanFauble, line 76: Settings for alert-manager to send out emails when configured in SigNoz

deployments/main.tf
```diff
@@ -31,24 +31,49 @@ module "dpe-sandbox-spacelift-development" {
...
+  # A public subnet is required for each AZ in which the worker nodes are deployed
+  public_subnet_cidrs                    = ["10.52.16.0/24", "10.52.17.0/24", "10.52.19.0/24"]
+  private_subnet_cidrs_eks_control_plane = ["10.52.18.0/28", "10.52.18.16/28"]
+  azs_eks_control_plane                  = ["us-east-1a", "us-east-1b"]
+
+  private_subnet_cidrs_eks_worker_nodes = ["10.52.28.0/22", "10.52.24.0/22", "10.52.20.0/22"]
+  azs_eks_worker_nodes                  = ["us-east-1c", "us-east-1b", "us-east-1a"]
+
+  enable_cluster_ingress = true
+  enable_otel_ingress    = true
```
BryanFauble, line 71: `enable_cluster_ingress` controls if envoy gateway is going to be deployed to the cluster. `enable_otel_ingress` controls whether or not a route to submit telemetry data is going to be created.

deployments/main.tf
```diff
@@ -31,24 +31,49 @@ module "dpe-sandbox-spacelift-development" {
...
+  ssl_hostname           = "dev.sagedpe.org"
+  auth0_jwks_uri         = "https://dev-sage-dpe.us.auth0.com/.well-known/jwks.json"
```
BryanFauble, line 73: Used for SSL termination at the gateway and to validate incoming requests on routes that require a JWT auth header be provided.

modules/apache-airflow/README.md
```diff
@@ -65,4 +65,20 @@ YAML
...
+1. Update the dockerfile within the orca-recipes repo
+2. Build the new image `docker build .`
+3. Tag the build image with the tag you want to use `docker tag sha256:... ghcr.io/sage-bionetworks-workflows/orca-recipes:0.0.1`
+4. Push to GCHR `docker push ghcr.io/sage-bionetworks-workflows/orca-recipes:0.0.1` (May require an admin of the repo to push this)
+5. Update the `values.yaml` file in this `modules/apache-airflow/templates` directory.
+
+Transitive dependencies may also need to be updated when building a new image for
+airflow, for example `py-orca` was updated in this example PR: <https://github.com/Sage-Bionetworks-Workflows/py-orca/pull/45>.
+Additionally, this PR covers what was completed in order to update the 
+requirements/dockerfile: <https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/71>.
```
BryanFauble, line 84: These are not new changes (Git is showing them in the diff for some reason). Please ignore changes here.

modules/apache-airflow/main.tf
```diff
@@ -66,7 +66,7 @@ spec:
   sources:
   - repoURL: 'https://airflow.apache.org'
     chart: airflow
-    targetRevision: 1.11.0
+    targetRevision: 1.15.0
```
BryanFauble, line 69: These are not new changes (Git is showing them in the diff for some reason). Please ignore changes here.

modules/apache-airflow/templates/values.yaml
```diff
@@ -25,6 +25,20 @@ fullnameOverride: ""
 # Provide a name to substitute for the name of the chart
 nameOverride: ""
 
+# Use standard naming for all resources using airflow.fullname template
```
BryanFauble, line 28: These are not new changes (Git is showing them in the diff for some reason). Please ignore changes here.

modules/victoria-metrics/templates/values.yaml
```diff
@@ -233,6 +233,9 @@ vmsingle:
       resources:
         requests:
           storage: 20Gi
+    nodeSelector: {
+      spotinst.io/node-lifecycle: "od"
+    }
```
BryanFauble, line 238: These are not new changes (Git is showing them in the diff for some reason). Please ignore changes here.

modules/signoz/main.tf
```diff
@@ -0,0 +1,114 @@
...
+            path: /metadata/namespace
+            value: ${var.gateway_namespace}
+          - op: replace
+            path: /spec/jwt/providers
+            value:
+              - name: auth0
+                remoteJWKS:
+                  uri: ${var.auth0_jwks_uri}
+                audiences:
+                  - ${var.cluster_name}-telemetry
```
BryanFauble, line 89: Adding this security policy let's us do filtering on this specific HTTP route for telemetry incoming to the cluster.

1. When a request enters the cluster this `SecurityPolicy` is evaluated since a request matches to the `http-route.yaml` defined in the `resources-otel-ingress` subdirectory. This allows us to test for one or more specific keys in the incoming JWT. We are verifying that audience is as expected


Testing the export a data with a valid JWT that does not have the expected audience fails with an HTTP 403:
`ERROR:opentelemetry.exporter.otlp.proto.http.trace_exporter:Failed to export batch code: 403, reason: Audiences in Jwt are not allowed`


@zaro0508 This is a key area of the security behind access to submit telemetry data to the cluster, and an area to pay close attention to.

In Auth0 I have an "API": https://auth0.com/docs/get-started/apis

> An API is an entity that represents an external resource, capable of accepting and responding to protected resource requests made by applications. In the [OAuth2 specification](https://tools.ietf.org/html/rfc6749), an API maps to the Resource Server.

For each JWT that is exchanged for this API it has a specific audience that I can filter for here.

**There is an alternative approach to this I was wanting feedback on:**

1. Instead of basing access on the "audience" portion of the JWT claim, I could instead base this on scopes: https://auth0.com/docs/get-started/apis/scopes#ways-to-use-scopes

> In an [API](https://auth0.com/docs/api), to implement access control. In this case, you need to define custom scopes for your API and then identify these scopes so that calling applications can use them.

2. Scopes would allow me to add a scope of `write:telemetry` to the JWT
3. By using scopes instead of audience I could use this as the source of authorization, instead of my current idea of using audience to handle this filtering.

BryanFauble: This was addressed as the latest version of envoy gateway supports scope based authorization via SecurityPolicies:

https://github.com/Sage-Bionetworks-Workflows/eks-stack/pull/48/files

PR #36, opened 2024-10-07
Envoy Gateway Minimum TLS

BryanFauble:
Problem:

1. Per Brad E (Sage) mentioned to me that for sending data encrypted in transit we must use the latest version of TLS (1.3)

Solution:

1. Apply the ClientTrafficPolicy (https://gateway.envoyproxy.io/docs/api/extension_types/#clienttrafficpolicy) resource to the Gateway to ensure all traffic is encrypted using TLS 1.3

Testing:

1. I deployed the resources out to the cluster and verified that TLS 1.2 fails, while TLS 1.3 is successful:

**Note: These URLs are no longer provisioned, so they cannot be used.**

`openssl s_client -connect a5dec82aabc5749ff8034e3a6167e08a-755132748.us-east-1.elb.amazonaws.com:443 -tls1_2`

Results in:
`40E71B0B787F0000:error:0A00042E:SSL routines:ssl3_read_bytes:tlsv1 alert protocol version:ssl/record/rec_layer_s3.c:907:SSL alert number 70`

image

`openssl s_client -connect a5dec82aabc5749ff8034e3a6167e08a-755132748.us-east-1.elb.amazonaws.com:443 -tls1_3`

Results in the self-signed cert used for testing:
```
verify error:num=18:self-signed certificate
verify return:1
depth=0 
verify return:1
---
Certificate chain
 0 s:
   i:
   a:PKEY: rsaEncryption, 2048 (bit); sigalg: RSA-SHA256
   v:NotBefore: Oct  7 17:02:55 2024 GMT; NotAfter: Jan  5 17:02:55 2025 GMT
---
Server certificate
-----BEGIN CERTIFICATE-----
.
.
.
```

### DIFF
--- a/deployments/stacks/dpe-k8s-deployments/main.tf
+++ b/deployments/stacks/dpe-k8s-deployments/main.tf
@@ -130,6 +130,6 @@ module "cluster-ingress" {
 
   # To determine more elegant ways to fill in these values, for example, if we have
   # a pre-defined DNS name for the cluster (https://sagebionetworks.jira.com/browse/IT-3931)
-  ssl_hostname        = "unknown-to-fill-in"
+  ssl_hostname        = "a5dec82aabc5749ff8034e3a6167e08a-755132748.us-east-1.elb.amazonaws.com"
   cluster_issuer_name = "selfsigned"
 }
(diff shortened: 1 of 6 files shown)


BWMac (approved):
👍🏻

PR #37, opened 2024-10-09
Signoz alert manager setup for SMTP

BryanFauble:
**Background documentation:**

1. https://signoz.io/docs/alerts-management/notification-channel/email/

**Problem:**

1. For alert manager to send out email it needs authentication to an SMTP server to send out those emails.

**Solution:**

1. I set up the connection for alert-manager to use gmail for SMTP

**Testing:**

1. I verified with temporary credentials that I was able send out a test notification:
image


BryanFauble:
Moved to https://github.com/Sage-Bionetworks-Workflows/eks-stack/pull/41

PR #38, opened 2024-10-17
Moving airflow to on-demand

BryanFauble:
**Problem:**

1. Redis goes down after around a month because the persistent volume gets deleted. My suspicion is that something is occurring with the spot instances that is causing issues.

**Solution:**

1. Move several pods to on-demand instances instead of spot instances to make sure they have sufficient up-time always, and prevent any issues with persistent volumes from detaching.

**Testing:**

1. Will be verifying in dev before applying to prod.

### DIFF
--- a/modules/apache-airflow/templates/values.yaml
+++ b/modules/apache-airflow/templates/values.yaml
@@ -108,7 +108,9 @@ images:
     pullPolicy: IfNotPresent
 
 # Select certain nodes for airflow pods.
-nodeSelector: {}
+nodeSelector: {
+  topology.kubernetes.io/zone: us-east-1a
+}
 affinity: {}
 tolerations: []
 topologySpreadConstraints: []
@@ -467,7 +469,7 @@ kerberos:
 # Airflow Worker Config
 workers:
   # Number of airflow celery workers in StatefulSet
-  replicas: 2
+  replicas: 1
   # Max number of old replicasets to retain
   revisionHistoryLimit: ~
 
@@ -636,7 +638,9 @@ workers:
   extraVolumeMounts: []
 
   # Select certain nodes for airflow worker pods.
-  nodeSelector: {}
+  nodeSelector: {
+    spotinst.io/node-lifecycle: "od"
+  }
   runtimeClassName: ~
   priorityClassName: ~
   affinity:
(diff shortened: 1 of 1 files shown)


modules/apache-airflow/templates/values.yaml
```diff
@@ -636,7 +638,9 @@ workers:
   extraVolumeMounts: []
 
   # Select certain nodes for airflow worker pods.
-  nodeSelector: {}
+  nodeSelector: {
+    spotinst.io/node-lifecycle: "od"
+  }
```
BryanFauble, line 643: This document describes where this label is coming from:
https://docs.spot.io/ocean/features/labels-and-taints?id=spotinstionode-lifecycle

modules/apache-airflow/templates/values.yaml
```diff
@@ -467,7 +469,7 @@ kerberos:
 # Airflow Worker Config
 workers:
   # Number of airflow celery workers in StatefulSet
-  replicas: 2
+  replicas: 1
```
rxu17, line 472: Curious why the number of workers got decreased?

rxu17 (approved):
LGTM!

modules/apache-airflow/templates/values.yaml
BryanFauble: I only reason I was originally using 2 replicas was because, Airflow support running in master-master mode, meaning database locks in postgres make sure data isn't processed twice. And if one of the replicas goes down it doesn't impact airflow.

The problem though is that redis out-of-the-box does not support multiple replicas so it was the single point of failure.

Since we're now running in on-demand instances instead of spot we're going to get the same reliability I was expecting to get out of 2 replicas.

modules/apache-airflow/templates/values.yaml
```diff
@@ -108,7 +108,9 @@ images:
     pullPolicy: IfNotPresent
 
 # Select certain nodes for airflow pods.
-nodeSelector: {}
+nodeSelector: {
+  topology.kubernetes.io/zone: us-east-1a
```
rxu17, line 112: Are not all types of airflow pods affected, therefore we just adjust to ondemand for the specific ones instead of here (it looks like this affects all airflow pods?)

BryanFauble: Not every type of airflow pod needs to be ran in an on-demand instance. For example the web UI can run on spot instances because if it goes down we just lose UI access for a short bit.

But if a worker or part of the infra of airflow goes down, it has a larger impact.

rxu17: Thanks for explaining!

BryanFauble:
Testing in the dev stack: I verified that on apply of the changes an on-demand node came up and all pods I wanted to were scheduled in the on-demand instance:
image


image

PR #39, opened 2024-10-18
[IBCDPE-1097] Shrink vpc subnets and move worker nodes

BryanFauble:
Initial comment chain on issue: https://github.com/Sage-Bionetworks-Workflows/eks-stack/pull/35#discussion_r1790712362

**Problem:**

1. Khai had some recommendations on VPC sizing for the EKS cluster
2. The EKS Best Practices Guide says:

> Kubernetes worker nodes can run in the cluster subnets, but it is not recommended. During [cluster upgrades](https://aws.github.io/aws-eks-best-practices/upgrades/#verify-available-ip-addresses) Amazon EKS provisions additional ENIs in the cluster subnets. When your cluster scales out, worker nodes and pods may consume the available IPs in the cluster subnet.

**Solution:**

1. Moving VPC subnets to `/20` subnet mask which is 4094 hosts
2. Moving the EKS control plane subnets (2 AZs) to their own subnet range
3. Moving the worker subnets (3 AZs) to their own subnet range

**Deployment:**

1. Since the initial VPC ranges cannot be shrunk we must remove the VPC and set it up again
2. Since the VPC of the EKS cluster cannot be moved into another VPC we must re-create the EKS cluster.

### DIFF
--- a/deployments/main.tf
+++ b/deployments/main.tf
@@ -37,10 +37,13 @@ module "dpe-sandbox-spacelift-development" {
   cluster_name = "dpe-k8-sandbox"
   vpc_name     = "dpe-sandbox"
 
-  vpc_cidr_block       = "10.51.0.0/16"
-  public_subnet_cidrs  = ["10.51.1.0/24", "10.51.2.0/24"]
-  private_subnet_cidrs = ["10.51.4.0/24", "10.51.5.0/24"]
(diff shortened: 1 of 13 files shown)


thomasyu888 (approved):
🔥 LGTM!  Going to pre-approve, but let's see @zaro0508's feedback here.

zaro0508 (approved):

PR #42, opened 2024-10-28
[IBCDPE-1111] upgrade airflow

BryanFauble:
**Problem:**

1. The current version of airflow is out of date and contains a number of bugs that are patched in newer versions that increase stability.

**Solution:**

1. Upgrading the helm chart and port over the values file into the updated values file for the chart.

**Testing:**

1. I deployed the old version of the chart to the kubernetes cluster and then verified that it was a seamless upgrade after the DB migrations ran.

### DIFF
--- a/modules/apache-airflow/templates/values.yaml
+++ b/modules/apache-airflow/templates/values.yaml
@@ -1997,7 +1997,7 @@ statsd:
   # Add custom annotations to the statsd configmap
   configMapAnnotations: {}
 
-  enabled: true
+  enabled: false
   # Max number of old replicasets to retain
   revisionHistoryLimit: ~
 
--- a/modules/apache-airflow/templates/values.yaml
+++ b/modules/apache-airflow/templates/values.yaml
@@ -70,7 +70,7 @@ defaultAirflowRepository: ghcr.io/sage-bionetworks-workflows/orca-recipes
 
 # TODO: Update this with out airflow repo/tag
 # Default airflow tag to deploy
-defaultAirflowTag: "development-0.0.2"
+defaultAirflowTag: "airflow-2.9.3-v1.0.0"
 
 # Default airflow digest. If specified, it takes precedence over tag
 defaultAirflowDigest: ~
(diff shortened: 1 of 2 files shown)


modules/apache-airflow/README.md
```diff
@@ -65,4 +65,20 @@ YAML
...
+the UI.
+
+## Building a new image for airflow
+The deployment of our airflow instance depends on a custom apache airflow image being 
+created and pushed to a public available GCHR url. The image is created from the
+`orca-recipes` git repo: <https://github.com/Sage-Bionetworks-Workflows/orca-recipes/tree/main>
+
+1. Update the dockerfile within the orca-recipes repo
+2. Build the new image `docker build .`
+3. Tag the build image with the tag you want to use `docker tag sha256:... ghcr.io/sage-bionetworks-workflows/orca-recipes:0.0.1`
```
thomasyu888, line 77: Ideally in the long term - a github action is set up to do this

modules/apache-airflow/templates/values.yaml
```diff
@@ -1750,7 +2081,7 @@ pgbouncer:
   command: ["pgbouncer", "-u", "nobody", "/etc/pgbouncer/pgbouncer.ini"]
   # Args to use for PgBouncer(templated).
   args: ~
-  auth_type: md5
+  auth_type: scram-sha-256
```
thomasyu888, line 2084: What is the significance of this change? Or is it the change in defaults

thomasyu888 (approved):
🔥 LGTM!  Pending Brad's return for final review

modules/apache-airflow/templates/values.yaml
BryanFauble: We are not using their version of pgbouncer so we are not affected. This was the default in the new chart version

modules/apache-airflow/README.md
BryanFauble: Agreed! It's not much work to do this manually, but a GH action in the `orca-recipes` repo will make the experience better. I created this ticket to add this logic: https://sagebionetworks.jira.com/browse/IBCDPE-1134

BWMac (approved):

PR #44, opened 2024-11-04
Signoz move to lets encrypt staging

BryanFauble:
**Problem:**

1. We were using self-signed SSL certs (A security issue) for TLS termination at the cluster gateway
2. We did not have a domain to point to

**Solution:**

1. Khai on the IT side created a domain for us to use (sagedpe.org) and he pointed (dev.sagedpe.org) using a CNAME to the load balancer serving traffic in the DNT DEV AWS account.
2. I am pointing the cluster issuer at let's encrypt.

**Testing:**

1. I verified that this setup was able to solve the ACME challenge and issue a cert: https://letsencrypt.org/docs/challenge-types/#http-01-challenge
image
image


BryanFauble:
Moved to https://github.com/Sage-Bionetworks-Workflows/eks-stack/pull/45

PR #45, opened 2024-11-04
[IBCDPE-1095] Signoz move to lets encrypt

BryanFauble:
**Problem:**

1. We were using self-signed SSL certs (A security issue) for TLS termination at the cluster gateway
2. We did not have a domain to point to

**Solution:**

1. Khai on the IT side created a domain for us to use (sagedpe.org) and he pointed (dev.sagedpe.org) using a CNAME to the load balancer serving traffic in the DNT DEV AWS account.
2. I am pointing the cluster issuer at let's encrypt.

**Testing:**

1. I verified that this setup was able to solve the ACME challenge and issue a cert: https://letsencrypt.org/docs/challenge-types/#http-01-challenge
image
image


modules/envoy-gateway/main.tf
```diff
@@ -63,15 +63,6 @@ spec:
...
             value: ${var.cluster_issuer_name}
-      - target:
-          kind: SecurityPolicy
-        patch: |-
-          - op: replace
-            path: /spec/jwt/providers
-            value:
-              - name: auth0
-                remoteJWKS:
-                  uri: ${var.auth0_jwks_uri}
```
BryanFauble, line 74: This was causing us issues to apply a blanket SecurityPolicy at the gateway to require JWT be provided due to the nature of how the ACME challenge needs to be solved:
https://letsencrypt.org/docs/challenge-types/#http-01-challenge

Because JWT was required the traffic to solve the challenge was being rejected due to HTTP 401.

Instead, we will expect that security policies will be individually applied to all HTTP Routes that are created for the cluster - **Like what is already in place for the telemetry endpoints**: [Like what is shown in this file](https://github.com/Sage-Bionetworks-Workflows/eks-stack/blob/ibcdpe-1095-signoz-move-to-lets-encrypt/modules/signoz/resources-otel-ingress/security-policy.yaml#L1-L12)

BryanFauble:
Merging this PR to collapse this mess of PR dependencies

PR #46, opened 2024-11-06
Upgrade helmchart for signoz

BryanFauble:
**Problem:**

1. SigNoz had a few version updates which include bug fixes/features.

**Solution:**

1. Upgrade the helm chart version
2. Port over the values file into the new values file.

**Testing:**

1. This was deployed to the dpe sandbox cluster to verify deployment

**Upgrade process:**

1. Update local helm repo `helm repo update` (See https://signoz.io/docs/install/kubernetes/others/ for setting up helm to begin with)
2. Show new latest version of `signoz/signoz` chart: `helm search repo signoz`
4. Create new `values.yaml` file that is used in the new chart version `helm show values signoz/signoz > values-new.yaml`
5. Perform a diff of the existing `values.yaml` file and the `values-new.yaml`
6. Port over any configuration that is needed, and adjust anything that is new
7. Commit/deploy the changes through the regular CI/CD process (ie: Pull request, spacelift deployment, argocd deployment)

### DIFF
--- a/modules/signoz/templates/values.yaml
+++ b/modules/signoz/templates/values.yaml
@@ -1301,12 +1301,6 @@ schemaMigrator:
   topologySpreadConstraints: []
 
   initContainers:
-		 
-			
-						   
-										   
-				 
-								
     init:
       enabled: true
       image:
@@ -1499,8 +1493,7 @@ otelCollector:
     type: ClusterIP
 
   # -- OtelCollector Deployment annotation.
-  annotations:
-							  
+  annotations: {}
   # -- OtelCollector pod(s) annotation.
   podAnnotations:
     signoz.io/scrape: 'true'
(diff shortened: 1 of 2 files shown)


modules/signoz/templates/values.yaml
```diff
@@ -1767,6 +1806,7 @@ otelCollector:
 
     autoscalingTemplate: []
     keda:
+      annotations:
```
BWMac, line 1809: is this value being empty intended?

BWMac (approved):

modules/signoz/templates/values.yaml
BryanFauble: It would probably cause issues. But also we are not using this, so install doesn't complain

PR #47, opened 2024-11-06
[SCHEMATIC-138] SigNoz cold storage and backups

BryanFauble:
https://sagebionetworks.jira.com/browse/SCHEMATIC-138

**Problem:**

1. We did not have anything in place to support tiered storage (Meaning some data on elastic block storage, and some data on S3) for cost effectiveness as we scale out with more data.
2. We had nothing in place to support backing up or restoring the clickhouse data.

**Solution:**

1. Implement an S3 module to support creating an S3 bucket along with the policies for an IAM role to manage that S3 bucket.
2. Using IRSA (IAM role for service account) to support attaching a kubernetes service account to assume an IAM role to get the permissions of that IAM role. This allows the service account to access and manage that S3 bucket.
3. Migrating to the usage of FluxCD instead of ArgoCD. FluxCD allows us to use `postRenders` and allows us to inject custom configuration into the rendered helm chart. This is a missing feature from ArgoCD and prevented us from continuing to use that tool.
4. Attaching the clickhouse-backup tool (From the same creators as the clickhouse operator being used) to support backups of the data and restore of the data.
5. Updating the storage configuration for the clickhouse server to support tiered storage and offloading "cold" data into S3. Lowering the overall amount of data that the EBS volumes need to contain.

**Testing:**

1. Brad and I tested the backup/restore process with a new k8s cluster
2. We verified data was being offloaded the EBS volumes into S3
3. Verified that the deployment of the resources were occurring through FluxCD


deployments/spacelift/dpe-k8s/main.tf
```diff
@@ -217,41 +217,44 @@ resource "spacelift_aws_integration_attachment" "k8s-deployments-aws-integration
...
-
-  depends_on = [
-    spacelift_space.dpe-space
-  ]
+# resource "spacelift_stack" "auth0" {
+#   count = 0
+#   github_enterprise {
+#     namespace = "Sage-Bionetworks-Workflows"
+#     id        = "sage-bionetworks-workflows-gh"
+#   }
```
BryanFauble, line 225: To be reverted before merge. We likely want to look to modify this eventually to conditionally create this `auth0` stack depending on if authentication is required for the stack. Currently it'll always be deployed, and, for the sandbox stack for Brad, we aren't going to be exposing the sandbox to the internet.

modules/flux-cd/templates/values.yaml
```diff
@@ -0,0 +1,327 @@
+# global
+
+installCRDs: true
+crds:
+  # -- Add annotations to all CRD resources, e.g. "helm.sh/resource-policy": keep
+  annotations: {}
```
BryanFauble, line 6: All default config. I didn't see anything needing to be updated

BWMac (approved):
LGTM! Thanks again for your help on this

PR #51, opened 2024-12-03
[SCHEMATIC-215] Enable ingress for SigNoz UI

BryanFauble:
https://sagebionetworks.jira.com/browse/SCHEMATIC-215

**Problem:**

1. SigNoz (Non-Enterprise) uses their own login mechanism that doesn't support SSO/SAML; We are reluctant to add JumpCloud SSO in-front of SigNoz as it will then require a 2-step login.
2. The Sage VPN is a viable option that requires users to be on the Sage VPN. As such we need to create a route to access the UI and restrict it to the public IP address.

**Solution:**

1. Creating a route in the Envoy Gateway to access the SigNoz UI.
2. Enable proxy protocol on the AWS load balancer to support forwarding the real client IP address via the `X-Forwarded-For` header.
3. Use an `authorization` SecurityPolicy to restrict access to the SigNoz HTTP route

**Testing:**

- [x] To deploy and test on k8s sandbox cluster
- I verified that the URL is accessible via the `dev` url:

image

### DIFF
--- a/modules/signoz/resources-otel-ingress/http-route.yaml
+++ b/modules/signoz/resources-otel-ingress/http-route.yaml
@@ -60,7 +60,7 @@ spec:
           weight: 1
       matches:
         - path:
-            type: PathPrefix
+            type: Exact
             value: /
         - headers:
             - name: Referer
--- a/modules/signoz/resources-otel-ingress/http-route.yaml
+++ b/modules/signoz/resources-otel-ingress/http-route.yaml
@@ -51,3 +51,17 @@ spec:
         - path:
             type: PathPrefix
             value: /telemetry/ui
+    - backendRefs:
+        - group: ""
+          kind: Service
+          name: signoz-frontend
+          namespace: signoz
+          port: 3301
+          weight: 1
+      matches:
+        - path:
+            type: PathPrefix
+            value: /
+        - headers:
+            - name: Referer
+              value: https://dev.sagedpe.org/telemetry/ui
--- a/modules/signoz/resources-otel-ingress/http-route.yaml
+++ b/modules/signoz/resources-otel-ingress/http-route.yaml
@@ -41,27 +41,7 @@ spec:
           namespace: signoz
           port: 3301
           weight: 1
-      filters:
-      - type: URLRewrite
-        urlRewrite:
-          path:
-            type: ReplacePrefixMatch
-            replacePrefixMatch: /
       matches:
         - path:
             type: PathPrefix
-            value: /telemetry/ui
-    - backendRefs:
-        - group: ""
-          kind: Service
-          name: signoz-frontend
-          namespace: signoz
-          port: 3301
-          weight: 1
-      matches:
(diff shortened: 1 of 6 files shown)


modules/signoz/resources-otel-ingress/http-route.yaml
```diff
@@ -24,3 +24,24 @@ spec:
...
+        - group: ""
+          kind: Service
+          name: signoz-frontend
+          namespace: signoz
+          port: 3301
+          weight: 1
+      matches:
+        - path:
+            type: PathPrefix
+            value: /
```
BryanFauble, line 47: SigNoz only supports working at the root URL of the domain it's running on. In favor of getting a solution I am going to use the ${environment}.sagedpe.org URLs. If we want to swap to using a different sub-domain later we'll be able to do that, for example https://github.com/Sage-Bionetworks-IT/organizations-infra/pull/1302 shows what we might do on the Sage IT infra stack.

On this side it's going to be a larger shift (And why I didn't want to worry about doing it now). It's because we'll need to make sure that we request SSL certs from Let's Encrypt and make sure it all works with the Listeners in the Envoy Gateway.

modules/signoz/resources-otel-ingress/security-policy.yaml
```diff
@@ -11,3 +11,22 @@ spec:
   jwt:
     providers: <replaced-by-kustomize>
   authorization: <replaced-by-kustomize>
+---
+apiVersion: gateway.envoyproxy.io/v1alpha1
```
zaro0508, line 15: alpha version?

modules/signoz/resources-otel-ingress/security-policy.yaml
```diff
@@ -11,3 +11,22 @@ spec:
...
+    kind: HTTPRoute
+    name: signoz-ui-route
+  authorization:
+    defaultAction: Deny
+    rules:
+    - action: Allow
+      principal:
+        clientCIDRs:
+        # Public IP address for the Sage VPN. `/32` CIDR mask means a single IP address.
+        - 52.44.61.21/32
```
zaro0508, line 32: i do't know about the rest but this looks like what we want.

zaro0508 (approved):

modules/signoz/resources-otel-ingress/security-policy.yaml
BryanFauble: It's the latest at the moment (https://gateway.envoyproxy.io/latest/api/extension_types/#gatewayenvoyproxyiov1alpha1)

The gateway itself is past the v1.x.x mark, but the APIs are subject to updates. I've been impressed with the changes this project has been publishing out and they have smoothed out a bunch of rough edges for shaping traffic into the K8s cluster. I've worked with other tools like Istio, and `Traefik Proxy` before & were a huge pain to manage at the time.

BWMac (approved):
LGTM!

PR #52, opened 2024-12-04
Add signoz dashboards

BryanFauble:
**Problem:**

1. SigNoz dashboards are a manual process and are not backed up during the backup of the telemetry data.

**Solution:**

1. Copy dashboards out into source control so we have tracking and backups of the content


BWMac (approved):
Dashboards look good!

PR #54, opened 2024-12-26
[DPE-1033/DPE-1188] Add docker authenticated pulls & Drop SigNoz self hosted

BryanFauble:
**Problem:**

1. Docker has some strict rate limits for pulling images when not logged in.
2. When a node goes down (Spot instance) or a cheaper on-demand node is spun up and workloads moved over many images need to be pulled in a short amount of time.
3. Self hosted SigNoz had a number of issues with the clickhouse database management

**Solution:**

1. Authenticating docker pulls to docker hub starting with SigNoz. 
2. Drop SigNoz from the K8s deployment

**Testing:**

1. Changes looked good in the sandbox cluster!
2. Will be tearing down SigNoz in all environments after merge


thomasyu888 (approved):
🔥 LGTM!

Thanks for doing this - we do have a sage organization which we could potentially add this individual.  Once added, this individual could have less restrictions with Docker pulls, but maybe we can figure that out later.

I'd like someone else on @Sage-Bionetworks-Workflows/dpe to do a final review.

Another critical note: be sure that Signoz cloud can be used before we drop this and then have to resurrect all of this.  Alberto did approve months ago but nothing is final until we get that subscription.

modules/apache-airflow/templates/values.yaml
```diff
@@ -2354,7 +2354,7 @@ redis:
 # Auth secret for a private registry
 # This is used if pulling airflow images from a private registry
 registry:
-  secretName: ~
+  secretName: docker-cfg
```
BWMac, line 2357: How would it work if we ever were to pull an image from another registry such as quay.io/would we ever need to do that?

BWMac (approved):
LGTM! Just one question

PR #57, opened 2025-03-14
[DPE-1234] Create S3 croissant bucket in DPE aws account

BryanFauble:
# **Problem:**

- A public S3 bucket is needed in the DPE account in order to handle the creation and access of croissant files created from Synapse Datasets. This [design doc describes](https://sagebionetworks.jira.com/wiki/spaces/DPE/pages/4006379521/DPE-1234+Automate+Synapse+Dataset+to+Croissant+JSON-LD+Files) the overall process for getting this data out of Snowflake and into these croissant files.

# **Solution:**

- Updating the s3 bucket module to support creating the bucket with public availability
- Add the creation of an S3 bucket to the DPE Prod stack so it will be created in the `org-sagebase-dpe-prod` AWS account

# **Testing:**

- I was able to verify in the DPE test AWS account that I could create an S3 bucket without public access, then after flipping the boolean, observe that the bucket could be listed by the public, and files could be read by the public.

### DIFF
--- a/deployments/stacks/dpe-k8s/main.tf
+++ b/deployments/stacks/dpe-k8s/main.tf
@@ -52,4 +52,5 @@ module "synapse_dataset_to_crossiant_metadata" {
   aws_account_id = var.aws_account_id
   cluster_name = var.cluster_name
   cluster_oidc_provider_arn = module.sage-aws-eks.cluster_oidc_provider_arn
+  public_access = true
 }
--- a/deployments/stacks/dpe-k8s/main.tf
+++ b/deployments/stacks/dpe-k8s/main.tf
@@ -47,7 +47,7 @@ module "sage-aws-ses" {
 module "synapse_dataset_to_crossiant_metadata" {
   count  = var.cluster_name == "dpe-k8" ? 1 : 0
   source      = "../../../modules/s3-bucket"
-  bucket_name = "synapse-croissant-metadata"
+  bucket_name = "synapse-croissant-metadata-2"
   enable_versioning = true
   aws_account_id = var.aws_account_id
   cluster_name = var.cluster_name
--- a/deployments/stacks/dpe-k8s/main.tf
+++ b/deployments/stacks/dpe-k8s/main.tf
@@ -43,3 +43,13 @@ module "sage-aws-ses" {
 
   email_identities = var.ses_email_identities
 }
+
+module "synapse_dataset_to_crossiant_metadata" {
+  count  = var.cluster_name == "dpe-k8-sandbox" ? 1 : 0
+  source      = "../../../modules/s3-bucket"
+  bucket_name = "synapse-croissant-metadata"
+  enable_versioning = true
+  aws_account_id = var.aws_account_id
+  cluster_name = var.cluster_name
+  cluster_oidc_provider_arn = module.sage-aws-eks.cluster_oidc_provider_arn
+}
--- a/deployments/stacks/dpe-k8s/main.tf
+++ b/deployments/stacks/dpe-k8s/main.tf
@@ -45,7 +45,7 @@ module "sage-aws-ses" {
 }
 
 module "synapse_dataset_to_crossiant_metadata" {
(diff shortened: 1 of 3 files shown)


deployments/stacks/dpe-k8s/main.tf
```diff
@@ -43,3 +43,15 @@ module "sage-aws-ses" {
...
+module "synapse_dataset_to_crossiant_metadata" {
+  # This points to the prod cluster - The only location where we want to create the bucket
+  count  = var.cluster_name == "dpe-k8" ? 1 : 0
+  source      = "../../../modules/s3-bucket"
+  bucket_name = "synapse-croissant-metadata"
+  enable_versioning = true
+  aws_account_id = var.aws_account_id
+  cluster_name = var.cluster_name
+  cluster_oidc_provider_arn = module.sage-aws-eks.cluster_oidc_provider_arn
+  public_access = false
```
BryanFauble, line 56: @ConsoleCatzirl @zaro0508 Before I flip on this `public_access` boolean flag do you mind giving these changes a review?

See [this design doc](https://sagebionetworks.jira.com/wiki/spaces/DPE/pages/4006379521/DPE-1234+Automate+Synapse+Dataset+to+Croissant+JSON-LD+Files) on what these changes are for.

I will be deploying this out to the `org-sagebase-dpe-prod` AWS account, and once these changes are reviewed I'll create the bucket and make it public.

This ties into changes [found here](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/77) which is the orchestration tool (Apache Airflow) that will run the python script to populate this S3 bucket.

BWMac (approved):
Changes LGTM, pre-approving pending review from IT folks 🚀

modules/s3-bucket/main.tf
```diff
@@ -8,6 +8,52 @@ resource "aws_s3_bucket" "bucket" {
...
+  bucket = aws_s3_bucket.bucket.id
+  acl    = "public-read"
+}
+
+resource "aws_s3_bucket_policy" "my_bucket_policy" {
+  count = var.public_access ? 1 : 0
+  bucket = aws_s3_bucket.bucket.bucket
+
+  policy = jsonencode({
+    Version = "2012-10-17",
```
linglp, line 45: I am also curious here why you pick this date for version number?

BryanFauble: https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_version.html

> 2012-10-17. This is the current version of the policy language, and you should always include a Version element and set it to 2012-10-17. Otherwise, you cannot use features such as [policy variables](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_variables.html) that were introduced with this version.

modules/s3-bucket/main.tf
```diff
@@ -8,6 +8,52 @@ resource "aws_s3_bucket" "bucket" {
...
+  rule {
+    object_ownership = "BucketOwnerPreferred"
+  }
+}
+
+resource "aws_s3_bucket_public_access_block" "access_block" {
+  count = var.public_access ? 1 : 0
+  bucket = aws_s3_bucket.bucket.id
+
+  block_public_acls       = false
```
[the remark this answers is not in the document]
linglp: Thanks Bryan. I mainly asked for learning purpose :)

zaro0508 (approved):
LGTM.  I'm not sure how many public buckets you plan to deploy but  just to note that IT use AWS securityhub to scan for public buckets.  When it finds one it will automatically generate a jira issue similar to https://sagebionetworks.jira.com/browse/IT-4072.  We then either need to change it to a private bucket or explicitly suppress the finding if is in fact meant to be public.  Please make a note to add this bucket to the [suppression list](https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/org-formation/075-security-hub/security-hub-suppress-infra.yaml#L401-L447)

PR #59, opened 2025-04-07
[DPE-1275] Use the latest orca-recipes image

BryanFauble:
# **Problem:**

- There is a newer orcs-recipes image

# **Solution:**

- Upgrade to the latest published image in order to include: https://github.com/Sage-Bionetworks-Workflows/py-orca/pull/50

# **Testing:**

- I verified that I could run these in docker compose in a codespaces environment


BWMac (approved):
👍🏻

PR #61, opened 2025-04-24
[DPE-1266] Improve responsiveness of webserver pod

jaymedina:
# **Problem:**

I'm seeing issues with lack of response either from the webserver or the scheduler for the Airflow namespace when trying to unpause and monitor Airflow DAGs:

image

When unpausing a DAG, it gets hung up and the connection to the port closes. Sometimes ending up in a perpetual restart loop:

image

# **Solution:**

Attempt to improve the responsiveness by:
- [x] upping the replica count
- [x] ensuring the webserver instances end up on different nodes
- [x] setting a memory and CPU resource allotment for requests

# **Testing:**

### DIFF
--- a/modules/apache-airflow/templates/values.yaml
+++ b/modules/apache-airflow/templates/values.yaml
@@ -681,14 +681,13 @@ workers:
     #   cpu: 100m
     #   memory: 128Mi
 
-
-  resources: {}
-  #  limits:
-  #   cpu: 100m
-  #   memory: 128Mi
-  #  requests:
-  #   cpu: 100m
-  #   memory: 128Mi
+  resources:
+    limits:
+     cpu: 100m
+     memory: 128Mi
+    requests:
+     cpu: 100m
+     memory: 128Mi
 
   # Grace period for tasks to finish after SIGTERM is sent from kubernetes
   terminationGracePeriodSeconds: 600
@@ -896,13 +895,13 @@ scheduler:
       maxUnavailable: 1
       # minAvailable: 1
 
-  resources: {}
-  #  limits:
-  #   cpu: 100m
-  #   memory: 128Mi
-  #  requests:
-  #   cpu: 100m
-  #   memory: 128Mi
+  resources:
+    limits:
+     cpu: 100m
+     memory: 128Mi
+    requests:
+     cpu: 100m
+     memory: 128Mi
 
   # This setting tells kubernetes that its ok to evict
   # when it wants to scale a node down.
@@ -966,12 +965,12 @@ scheduler:
     # Number of days to retain logs
     retentionDays: 15
     resources: {}
-    #  limits:
-    #   cpu: 100m
-    #   memory: 128Mi
-    #  requests:
-    #   cpu: 100m
-    #   memory: 128Mi
+      limits:
+       cpu: 100m
+       memory: 128Mi
+      requests:
+       cpu: 100m
+       memory: 128Mi
     # Detailed default security context for logGroomerSidecar for container level
     securityContexts:
       container: {}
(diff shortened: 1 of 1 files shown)


modules/apache-airflow/templates/values.yaml
```diff
@@ -1474,14 +1474,14 @@ webserver:
   priorityClassName: ~
   affinity: {}
```
BryanFauble, line 1475: Delete this empty `{}`

modules/apache-airflow/templates/values.yaml
```diff
@@ -1472,16 +1472,15 @@ webserver:
   # Select certain nodes for airflow webserver pods.
   nodeSelector: {}
   priorityClassName: ~
-  affinity: {}
```
BryanFauble, line 1475: `affinity:` should be kept

jaymedina:
We should also increase memory cap? It's at 70%
image

BryanFauble:
> We should also increase memory cap? It's at 70% image

Nothing in the airflow configs is capping memory. What could be set is the "requests" portion:

https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#requests-and-limits


To ensure that the pods "Have at least X memory"

The reason why that message for memory level shows up is because of the "Bin packing problem"

https://spot.io/blog/beyond-savings-overlooked-aspects-of-container-optimization/

Spot io, our auto scaler, is packing workloads into ec2 instance to reduce cost.



While we are here and making changes why don't we go ahead and set the requests (Just leave the default values airflow maintainer provided) on all workloads we have enabled. It would be deleting the empty `{}` and uncommenting the requests portion with cpu/memory. Here is one the spots we would update:

https://github.com/Sage-Bionetworks-Workflows/eks-stack/blob/dbd4fc57c23705891e1b57343d416811d1fb81b0/modules/apache-airflow/templates/values.yaml#L899

modules/apache-airflow/templates/values.yaml
```diff
@@ -681,14 +681,13 @@ workers:
...
-
-  resources: {}
-  #  limits:
-  #   cpu: 100m
-  #   memory: 128Mi
-  #  requests:
-  #   cpu: 100m
-  #   memory: 128Mi
+  resources:
+    limits:
```
BryanFauble, line 685: Let's leave limits on all of these commented out. I'd rather we only set the requests

jaymedina:
Thanks for your help @BryanFauble! Let me know if there's anything you'd like me to fold into this PR while it's still open.

modules/apache-airflow/templates/values.yaml
```diff
@@ -677,18 +677,17 @@ workers:
     #  limits:
     #   cpu: 100m
     #   memory: 128Mi
-    #  requests:
-    #   cpu: 100m
-    #   memory: 128Mi
+      requests:
```
BryanFauble, line 680: Empty `{}` for resources needs to be deleted

jaymedina: Where is the `{}`?

jaymedina: NVM!

modules/apache-airflow/templates/values.yaml
```diff
@@ -2446,12 +2445,12 @@ cleanup:
   labels: {}
 
   resources:
-   limits:
-    cpu: 1000m
-    memory: 512Mi
-   requests:
-    cpu: 100m
-    memory: 128Mi
+  #  limits:
```
BryanFauble, line 2448: For spots where it is `enabled: false`, no changes need to be made as it won't have any affect on what we have deployed.

PR #62, opened 2025-04-30
[SYNPY-1587] Adds AWS API Gateway & SQS modules

BWMac:
# **Problem:**

We do not currently have reusable IaC components to spin up SQS queues.

# **Solution:**

Implemented Terraform modules to create an AWS API Gateway and SQS queues that integrate with a specified API Gateway. With both of these components, we can send requests to the API Gateway and route them to appropriate SQS queues which can then be polled by downstream processes. This implementation allows us to have multiple SQS queues per API Gateway by simply assigning one route to each queue. This is just one way we could take a polling approach to receive messages generated by Synapse Webhooks. Another way we might want to look into is a push-based approach with SNS topics that downstream resources subscribe to. 

For the purposes of SYNPY-1587, I created an [Airflow DAG](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/87) that polls the SQS queue, but lighter-weight AWS resources like Lambda functions with SQS triggers or that subscribe to SNS topics (also managed with IaC code) may be preferable for production notification systems.

# **Testing:**

Deployed the infrastructure on the dnt-dev AWS account using Terraform + Spacelift and tested with a [Python script](https://gist.github.com/BWMac/fc406be9e4a2aee39c659465e9c9f3e7).

# **Notes:**

When merged, this PR will not include any API Gateways or SQS queues being deployed to the production account because we do not have immediate need for these resources.

### DIFF
--- a/deployments/stacks/dpe-k8s/main.tf
+++ b/deployments/stacks/dpe-k8s/main.tf
@@ -57,15 +57,13 @@ module "synapse_dataset_to_crossiant_metadata" {
   enable_cors = true
 }
 
-module "synapse-webhook-test" {
+module "synapse-sqs-test" {
   source = "../../../modules/aws-sqs"
   environment = var.cluster_name == "dpe-k8s-sandbox" ? "sandbox" : "dev"
-  name = "synapse-webhook"
-  queue_name = "synapse-webhook-queue"
-  namespace = "sqs-test"
+  name = "synapse-sqs-test"
   aws_account_id = var.aws_account_id
   aws_region = var.region
-  ack_controller_role_arn = module.sage-aws-eks.cluster_oidc_provider_arn
+  cluster_oidc_provider_arn = module.sage-aws-eks.cluster_oidc_provider_arn
   
   tags = {
     Environment = var.cluster_name
--- a/deployments/stacks/dpe-k8s/main.tf
+++ b/deployments/stacks/dpe-k8s/main.tf
@@ -57,17 +57,34 @@ module "synapse_dataset_to_crossiant_metadata" {
   enable_cors = true
 }
 
-module "synapse-sqs-test" {
+module "synapse-webhook-api-gateway" {
+  source = "../../../modules/aws-api-gateway"
+  environment = var.cluster_name == "dpe-k8s-sandbox" ? "sandbox" : "dev"
+  name = "synapse-webhook-api-gateway"
+}
+
+module "synapse-sqs-create-queue" {
   source = "../../../modules/aws-sqs"
   environment = var.cluster_name == "dpe-k8s-sandbox" ? "sandbox" : "dev"
-  name = "synapse-sqs-test"
+  name = "synapse-sqs-create-queue"
   aws_account_id = var.aws_account_id
   aws_region = var.region
   cluster_oidc_provider_arn = module.sage-aws-eks.cluster_oidc_provider_arn
   
-  tags = {
-    Environment = var.cluster_name
-    Application = "synapse"
-    ManagedBy   = "terraform"
-  }
+  # API Gateway integration
+  api_gateway_id = module.api_gateway.api_id
+  route_path = "/create"
+}
+
+module "synapse-sqs-delete-queue" {
+  source = "../../../modules/aws-sqs"
+  environment = var.cluster_name == "dpe-k8s-sandbox" ? "sandbox" : "dev"
+  name = "synapse-sqs-delete-queue"
+  aws_account_id = var.aws_account_id
+  aws_region = var.region
+  cluster_oidc_provider_arn = module.sage-aws-eks.cluster_oidc_provider_arn
+  
+  # API Gateway integration
+  api_gateway_id = module.api_gateway.api_id
+  route_path = "/delete"
 }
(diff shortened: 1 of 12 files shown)


deployments/stacks/dpe-k8s/main.tf
```diff
@@ -56,3 +56,37 @@ module "synapse_dataset_to_crossiant_metadata" {
...
+  name = "synapse-sqs-delete-queue"
+  aws_account_id = var.aws_account_id
+  aws_region = var.region
+  cluster_oidc_provider_arn = module.sage-aws-eks.cluster_oidc_provider_arn
+  
+  # API Gateway integration
+  enable_api_gateway_integration = true
+  api_gateway_id = module.synapse-webhook-api-gateway.api_id
+  route_path = "/delete"
+}
```
BWMac, line 92: We don't have a current business need for these SQS queues so I will be deleting these module invocations before merging, but wanted to leave this in as an example to show how I deployed the gateway and queues in the sandbox for testing

thomasyu888: Once we do the writeup, we can send this out to scidata and ask for some use cases.  Does the queue only work with one project?

thomasyu888 (commented):
🔥 This is great - agreed on not deploying anything in prod until we absolutely need it. That said, id love to demo this during sprint review or something.

deployments/stacks/dpe-k8s/main.tf
BWMac: Yes, each Webhook can only subscribe to one Synapse Entity. You can point multiple Webhooks at a single SQS queue though so you could monitor multiple entities with one queue. The only limit there is 25 Webhooks per user.

modules/aws-api-gateway/main.tf
```diff
@@ -0,0 +1,11 @@
+resource "aws_apigatewayv2_api" "api_gateway" {
+  name          = "${var.environment}-${var.name}-api"
+  protocol_type = "HTTP"
+  description   = "API Gateway for various integrations"
+}
+
+resource "aws_apigatewayv2_stage" "api_gateway_stage" {
+  api_id = aws_apigatewayv2_api.api_gateway.id
+  name   = "$default"
```
BryanFauble, line 9: Should name be coming in as a variable? I've never seen this `"$default"` syntax before, do you have docs on what this is?

modules/aws-api-gateway/variables.tf
```diff
@@ -0,0 +1,15 @@
...
+variable "name" {
+  description = "The name prefix for resources created by this module"
+  type        = string
+}
+
+variable "tags" {
+  description = "A map of tags to assign to the API Gateway"
+  type        = map(string)
+  default     = {}
+}
```
BryanFauble, line 15: I've been defaulting this to the no cost center:

https://github.com/Sage-Bionetworks-Workflows/eks-stack/blob/db9e56b88118c168567163eae19523a70653afa2/modules/s3-bucket/variables.tf#L6-L12

modules/aws-sqs/README.md
```diff
@@ -0,0 +1,170 @@
+# AWS SQS Module
+
+This module creates an SQS queue with optional API Gateway integration.
+
+## Usage
```
BryanFauble, line 5: This might be a good spot to document the additional changes that might be needed to get this to work with the airflow DAG. [Comment from that PR](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/87):

> I needed to give Airflow access to SQS in the dnt-dev AWS account so I created the airflow-sqs-access IAM user with the AmazonSQSFullAccess IAM role attached, and an AWS Secrets connection string airflow/connections/AWS_DNT_DEV_SQS_CONN in the dpe-prod account. The SQS infrastructure would need to be deployed in dpe-prod and similar access would need to be granted if we wanted to use this fully in production.

There is a bit of tight coupling that whomever uses this for the airflow dag would need to be aware of.

modules/aws-sqs/README.md
```diff
@@ -0,0 +1,170 @@
...
+module "api_gateway" {
+  source      = "path/to/modules/aws-api-gateway"
+  environment = "dev"
+  name        = "events-gateway"
+  
+  tags = {
+    Environment = "dev"
+    Application = "api-gateway"
+    ManagedBy   = "terraform"
+  }
```
BryanFauble, line 20: cost center tags through the module

modules/aws-sqs/variables.tf
```diff
@@ -0,0 +1,73 @@
...
+  description = "The ID of the API Gateway to integrate with. Required if enable_api_gateway_integration is true."
+  type        = string
+  default     = ""
+}
+
+variable "route_path" {
+  description = "The path for the API Gateway route (e.g., /events, /messages)"
+  type        = string
+  default     = "/events"
+}
```
BryanFauble, line 73: Add a tags variable

BryanFauble (changes requested):
I left a handful of comments around the code to patch up, but overall this is great work! Awesome to see more terraform scripts.

Since there is a little cleanup to remove the deployed changes from the DNT dev account, shoot me a new review when those get cleaned up and I'll approve.

BryanFauble (approved):
LGTM!

PR #63, opened 2025-05-12
[DPE-1312] Updates Airflow container version

BWMac:
# **Problem:**

We need to use the new container version from `orca-recipes`.

# **Solution:**

Updated the container version.

# **Testing:**

N/A

### DIFF
--- a/modules/apache-airflow/templates/values.yaml
+++ b/modules/apache-airflow/templates/values.yaml
@@ -68,7 +68,7 @@ airflowHome: /opt/airflow
 defaultAirflowRepository: ghcr.io/sage-bionetworks-workflows/orca-recipes
 
 # Default airflow tag to deploy
-defaultAirflowTag: "0.2.1"
+defaultAirflowTag: "latest"
 
 # Default airflow digest. If specified, it takes precedence over tag
 defaultAirflowDigest: ~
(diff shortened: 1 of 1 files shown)


modules/apache-airflow/templates/values.yaml
```diff
@@ -68,7 +68,7 @@ airflowHome: /opt/airflow
 defaultAirflowRepository: ghcr.io/sage-bionetworks-workflows/orca-recipes
 
 # Default airflow tag to deploy
-defaultAirflowTag: "0.2.1"
+defaultAirflowTag: "latest"
```
BryanFauble, line 71: I don't think that we should use "latest" here. I would rather that we be very intentional about what could be and what is deployed to the K8s cluster. I do not want us to run into a situation where "latest" is updated and portions of the Airflow cluster don't get updated. This can happen if pods get moved around to new nodes, where the image pull process would get the new "latest" image and be more up to date than other pods in the cluster.

Instead, we should use a specific version tag to ensure consistent deployments across the entire cluster.

BWMac: Makes sense. Just made that update.

BryanFauble (approved):
LGTM!

When you sync the change in ArgoCD remember to select and sync only those components which have changes

PR #66, opened 2025-07-18
Increase storage size from 8Gi to 30Gi in cluster configuration

BryanFauble:
# **Problem:**

- The Postgres DB ran out of storage

# **Solution:**

- Increase to 30Gi

# **Testing:**

- Will be testing out the deployment


linglp (approved):
LGTM! Thanks for the fix

PR #68, opened 2025-07-24
[feat]: upgrade K8s cluster to 1.32

linglp:
# **Problem:**
The current K8s cluster is reaching the end of life by October this year. See documentation [here](https://kubernetes.io/releases/#release-v1-31). 

# **Solution:**
Update to use K8s version 1.32 for now and 1.33 in the end [here](https://kubernetes.io/releases/#release-v1-33)

PR #69, opened 2025-07-25
[feat]: update version to 1.33

linglp:
# **Problem:**
Related to: https://github.com/Sage-Bionetworks-Workflows/eks-stack/pull/68
Updated version to 1.33.


BryanFauble (approved):
Worked with @linglp to apply the updates to the staging and prod clusters - and rolled the clusters so all nodes went to 1.32, we should be ok to move to 1.33

PR #72, opened 2025-08-11
[DPE-1423] Swap ocean-aws-k8s module to support provided commit

BryanFauble:
# **Problem:**

- 1.33 broke the Spotinst terraform module as the filter used is no longer valid for AWS published images

# **Solution:**

- Use the commit hash provided by support

# **Testing:**

- Update source and point the stack to use the local version instead of one from the private module registery


thomasyu888 (approved):
LGTM!

PR #84, opened 2026-04-03
Update snowflake_external_id in main.tf

jaymedina:
# **Problem:**

The external ID for the `dev` storage integration was also outdated

# **Solution:**

Update accordingly

# **Testing:**

Attempting to refresh the stage pointing to this storage integration should work:

image

Looks good ✅

### DIFF
--- a/deployments/main.tf
+++ b/deployments/main.tf
@@ -166,7 +166,7 @@ module "snowflake-spacelift-development" {
   
   # Snowflake authentication
   snowflake_principal_arn = "arn:aws:iam::365909334157:user/m2nb0000-s"
-  snowflake_external_id   = "UO70315_SFCRole=2_0VqqeWxjpuUoM4zVlbsAl5vnp/o="
+  snowflake_external_id   = "UO70315_SFCRole=2_PQyc1mUCluvv6W2lycbjCbYbjjI="
 }
 
 module "snowflake-spacelift-production" {
--- a/deployments/main.tf
+++ b/deployments/main.tf
@@ -166,7 +166,7 @@ module "snowflake-spacelift-development" {
   
   # Snowflake authentication
   snowflake_principal_arn = "arn:aws:iam::365909334157:user/m2nb0000-s"
-  snowflake_external_id   = "UO70315_SFCRole=2_PQyc1mUCluvv6W2lycbjCbYbjjI="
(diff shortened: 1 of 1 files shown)


thomasyu888 (commented):
🔥 LGTM - I'll defer to Phil for final review!

philerooski (commented):
These are not the `STORAGE_AWS_EXTERNAL_ID` values I see for the`SYNAPSE_SNAPSHOTS_DEV` or `SYNAPSE_SNAPSHOTS_PROD` storage integrations. Am I misunderstanding something?

jaymedina:
@philerooski I found the issue:

The `CREATE OR REPLACE` DDL was being used to generate the storage integrations, which does 2 things we were seeing issues with ([doc](https://docs.snowflake.com/en/sql-reference/sql/create-storage-integration#usage-notes)):
1. Break the association between the storage integration and its external stage because a new hidden ID was generated alongside the storage integration
2. Regenerates the storage integration's external ID which is what I use to grant the integration access to our IAM role for the DPE S3 bucket (which is the error you're pointing out). We don't hardcode this external ID so it gets regenerated every time the integration gets re-created

And because the storage integration creation is performed in the `admin/integrations.sql` script which runs for every deployment to `main` ([see here](https://github.com/Sage-Bionetworks/snowflake/blob/56345025d11492239a9c2303bcbcf9da49028c9e/.github/workflows/ci.yaml#L185)), this integration kept getting re-generated, causing the 2 issues above to resurface.

I'll make an update to `integrations.sql` and update the external ID on this branch to re-request a review. Thanks for the catch!

jaymedina:
I can confirm that after the most recent run on [snowsql_admin](https://github.com/Sage-Bionetworks/snowflake/actions/runs/24204938049/job/70658463067), the storage integration external IDs have remained [the same](https://github.com/Sage-Bionetworks-Workflows/eks-stack/pull/84/changes/bb4af241c2a9584903dc78fa38253e4d28bde025), and therefore still connected with the original stage.

I will now redeploy our RDS snapshot stacks on spacelift...

`prod` stage looks good ✅
image

`dev` stage looks good ✅ (that is, no error message)
image

Re-requesting a review so folks can triple check the latest external IDs

thomasyu888 (commented):
🔥 LGTM, I'll defer to @philerooski for final review

philerooski (approved):

PR #88, opened 2026-07-14
[IT-4153] Add airflow-croissant IRSA role to replace static access key

BryanFauble:
# **Problem:**

Prod Airflow reaches the `synapse-croissant-metadata[-minimal]` S3 buckets and the `airflow/*` Secrets Manager secrets using long-lived access keys on the `airflow-secrets-backend` IAM user. The S3 key is stored *inside* the `AWS_SYNAPSE_CROISSANT_METADATA_S3_CONN` connection secret. These keys must be manually rotated every 90 days to satisfy CIS AWS Foundations Benchmark 1.14 ([IT-4153](https://sagebionetworks.jira.com/browse/IT-4153)), which recurs indefinitely.

Verified current state:
- Airflow pods have **no** IRSA annotation / Pod Identity today; they reach AWS via the node instance role `work_profile_iam_role_dpe-k8` (`SecretsManagerReadWrite`) for secret resolution, and the static key (from the connection secret) for croissant S3.
- Both croissant DAGs use the single connection `AWS_SYNAPSE_CROISSANT_METADATA_S3_CONN` but write to **two** buckets.
- DAG parsing calls `Variable.get(...)` at module scope (`synapse-dataset-to-croissant.py`), so the **scheduler** (and webserver, which imports DAGs) also read Secrets Manager — not just the worker.

# **Solution:**

Introduce IRSA so prod Airflow assumes an IAM role via the cluster OIDC provider instead of using a static key.

- **New `airflow-croissant` IAM role** (`modules/apache-airflow/iam.tf`), gated to the prod cluster (`cluster_name == "dpe-k8"`):
  - **Trust** scoped to an explicit list of the four service accounts — `system:serviceaccount:airflow:airflow-{worker,scheduler,triggerer,webserver}` — via `sub`/`aud` conditions on the OIDC provider (closes the open-trust gap the existing single-bucket croissant roles have).
  - **Policy is a superset of the pods' current ambient access** so nothing loses access: S3 `Put/Get/Delete/ListBucket` on **both** croissant buckets (mirrors the inline `croissant-crud-for-airflow` user policy) **plus** `secretsmanager:GetSecretValue` on `airflow/connections/*` and `airflow/variables/*`, and `ListSecrets`.
- **Service account annotations** injected via **prod-only inline `helm.values`** in the ArgoCD Application (`modules/apache-airflow/main.tf`), rather than the git-synced `values.yaml` which is shared with staging. Staging (same account, separate cluster/OIDC) renders no annotation and keeps its ambient node-role behavior.
- Wired `cluster_name` / `aws_account_id` / `cluster_oidc_provider_arn` into the module from `dpe-k8s-deployments`.

Secrets Manager scope is intentionally the full `airflow/*` prefixes (not just croissant): once the worker is annotated, resolving **every** connection (e.g. `AWS_TOWER_PROD_S3_CONN`) goes through the role, so a narrower scope would break unrelated DAGs.

No KMS grant is needed (all `airflow/*` secrets use the AWS-managed default key) and no `PutObjectAcl` (the croissant `S3Hook` only sets `ContentType`; public-read is bucket-policy based).

# **Testing:**

- `tofu fmt` clean; `tofu validate` passes on the module.
- Rendered the ArgoCD Application template for both clusters: prod emits the inline `values:` block annotating all four SAs with the correct role ARN; staging emits no block. Confirmed the block parses as valid YAML into the structure the Airflow chart expects.
- Verified the OIDC-URL derivation against the real provider ARN produces the correct `...:sub` / `...:aud` condition keys.

## Cutover plan (perform in order — zero-gap, reversible)

> The static key stays valid until step 3, so there is always a working path to S3.

1. **Merge this PR.** ArgoCD syncs; **roll** `worker`, `scheduler`, `triggerer`, `webserver` so the SA token/annotation is injected.
2. **Verify the overlap** while the key is *still* in the secret:
   - On a fresh worker pod: `env | grep AWS_` shows `AWS_ROLE_ARN` + `AWS_WEB_IDENTITY_TOKEN_FILE`.
   - From that pod: `aws s3 ls s3://synapse-croissant-metadata` and `s3://synapse-croissant-metadata-minimal` succeed (via IRSA).
   - Confirm the scheduler parses a croissant DAG **and** a Tower/challenge DAG (`AWS_TOWER_PROD_S3_CONN`) with no Secrets Manager errors.
3. **Strip the static creds** from the `AWS_SYNAPSE_CROISSANT_METADATA_S3_CONN` secret (manual; operationally managed):
   `{"conn_type":"aws","login":"AKIA…","password":"…"}` → `{"conn_type":"aws","region_name":"us-east-1"}`.
4. **Validate end-to-end:** trigger both croissant DAGs; confirm S3 write/list/delete succeed and the public object URL serves. Rollback if needed = re-add the key to the secret.

## Follow-up (separate, after both this and the orca-recipes SSO PR are verified live)
- Confirm via `aws iam get-access-key-last-used` the S3 key has gone idle, then delete both access keys, the `airflow-secrets-backend` user, and its policies; resolve IT-4153. (Companion PR: developer SSO migration in `orca-recipes`.)

[IT-4153]: https://sagebionetworks.jira.com/browse/IT-4153?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/apache-airflow/main.tf
+++ b/modules/apache-airflow/main.tf
@@ -71,6 +71,25 @@ spec:
       releaseName: airflow
       valueFiles:
       - $values/modules/apache-airflow/templates/values.yaml
+%{if local.create_croissant_irsa}
+      values: |
+        workers:
+          serviceAccount:
+            annotations:
+              eks.amazonaws.com/role-arn: ${local.croissant_role_arn}
+        scheduler:
+          serviceAccount:
+            annotations:
+              eks.amazonaws.com/role-arn: ${local.croissant_role_arn}
+        triggerer:
+          serviceAccount:
+            annotations:
+              eks.amazonaws.com/role-arn: ${local.croissant_role_arn}
+        webserver:
+          serviceAccount:
+            annotations:
+              eks.amazonaws.com/role-arn: ${local.croissant_role_arn}
+%{endif}
   - repoURL: 'https://github.com/Sage-Bionetworks-Workflows/eks-stack.git'
     targetRevision: ${var.git_revision}
     ref: values
--- a/modules/apache-airflow/main.tf
+++ b/modules/apache-airflow/main.tf
@@ -71,24 +71,24 @@ spec:
       releaseName: airflow
       valueFiles:
       - $values/modules/apache-airflow/templates/values.yaml
-%{if local.create_croissant_irsa}
+%{if local.create_airflow_irsa}
       values: |
         workers:
           serviceAccount:
             annotations:
-              eks.amazonaws.com/role-arn: ${local.croissant_role_arn}
+              eks.amazonaws.com/role-arn: ${local.airflow_irsa_role_arn}
         scheduler:
           serviceAccount:
             annotations:
-              eks.amazonaws.com/role-arn: ${local.croissant_role_arn}
+              eks.amazonaws.com/role-arn: ${local.airflow_irsa_role_arn}
         triggerer:
           serviceAccount:
             annotations:
-              eks.amazonaws.com/role-arn: ${local.croissant_role_arn}
+              eks.amazonaws.com/role-arn: ${local.airflow_irsa_role_arn}
         webserver:
           serviceAccount:
             annotations:
-              eks.amazonaws.com/role-arn: ${local.croissant_role_arn}
+              eks.amazonaws.com/role-arn: ${local.airflow_irsa_role_arn}
 %{endif}
   - repoURL: 'https://github.com/Sage-Bionetworks-Workflows/eks-stack.git'
     targetRevision: ${var.git_revision}
--- a/deployments/stacks/dpe-k8s-deployments/main.tf
+++ b/deployments/stacks/dpe-k8s-deployments/main.tf
@@ -59,12 +59,15 @@ module "airflow" {
   depends_on = [module.victoria-metrics, module.argo-cd]
   # source       = "spacelift.io/sagebionetworks/airflow/aws"
   # version      = "0.4.0"
-  source              = "../../../modules/apache-airflow"
-  auto_deploy         = var.auto_deploy
-  auto_prune          = var.auto_prune
-  git_revision        = local.git_revision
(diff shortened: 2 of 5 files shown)


thomasyu888 (commented):
🔥 This makes sense to me and thanks for your work here!

1. An AWS IAM role with an attached policy that has the permissions to do basic CRUD operations within objects within a bucket.  Not sure if we use S3 object versions so this may not be needed

```
"s3:GetObjectVersion",
"s3:DeleteObjectVersion"
```

2. The doc here: https://docs.aws.amazon.com/eks/latest/userguide/associate-service-account-role.html to annotate the mapping which lines up with `eks.amazonaws.com/role-arn: croissant_role_arn`

I'll defer to @rxu17 and @linglp for a final review.  It may be good to step through some of this live given context around this ticket https://sagebionetworks.jira.com/browse/DPE-1652

modules/apache-airflow/main.tf
rxu17, line 1: In the future if we add diff stacks for deployment (like croissant being its own type of stack?), do they need to go through this process here like croissant is? Or rather why is croissant specifically affected and is that something we need to watch out for in the future with other additions here?

BryanFauble: Croissant is specifically affected here because it needs additional permissions to read/write/delete data from an S3 bucket. Any time that permissions are concerned this is something we would need to update so that the airflow processes are permitted to perform those actions.

rxu17: Gotcha! Could this be added to the docs somewhere?

BryanFauble: Added! I also realize that I should have made the names of the roles generic since it could include additional permissions - Not just croissant, so I also did that change too.

rxu17 (approved):
LGTM!

